PR #4, opened 2017-05-20
Fix Typos in Readme

jgeiger:
The word referrer was also incorrect, but it may need a code change.

Not sure if hlog has this method, but the one previously listed is spelled wrong.
 c = c.Append(hlog.ReferrerHandler("referrer"))


rs:
Thx for the PR. Referer is an historic typo: https://en.m.wikipedia.org/wiki/HTTP_referer.

Could you please revert that part of the commit so I can merge the other fixes?

jgeiger:
Sorry I did a live edit so it didn't squash the commits...

PR #6, opened 2017-06-17
Optimistically expect simple strings for json

tidwall:
Performance update to appendJSONString so that it now checks if the input is a simple string that contains no double quotes, control characters, or unicode. If simple then the operation is only three appends. [double-quote, string, double-quote].

If a non-simple character is encountered then all of the previous characters are appended and the operation falls back to the original method for the remaining characters.

Before:

```
BenchmarkLogEmpty-8	    100000000	        17.1 ns/op
BenchmarkDisabled-8	    500000000	         4.12 ns/op
BenchmarkInfo-8		    20000000	       101 ns/op
BenchmarkContextFields-8    20000000	       105 ns/op
BenchmarkLogFields-8	     5000000	       281 ns/op
```

After:

```
BenchmarkLogEmpty-8	    100000000	        16.7 ns/op
BenchmarkDisabled-8	    500000000	         3.79 ns/op
BenchmarkInfo-8		    30000000	        44.8 ns/op
BenchmarkContextFields-8    30000000	        67.5 ns/op
BenchmarkLogFields-8	    10000000	       197 ns/op
```

### DIFF
--- a/json.go
+++ b/json.go
@@ -4,10 +4,28 @@ import "unicode/utf8"
 
 const hex = "0123456789abcdef"
 
+// appendJSONString encodes the input string to json and appends
+// the encoded string to the input byte slice.
+//
+// The operation loops though each byte in the string looking
+// for characters that need json or utf8 encoding. If the string
+// does not need encoding, then the string is appended in it's
+// entirety to the byte slice.
+// If we encounter a byte that does need encoding, switch up
+// the operation and perform a byte-by-byte read-encode-append.
 func appendJSONString(dst []byte, s string) []byte {
+	// Start with a double quote.
 	dst = append(dst, '"')
+	// Loop through each character in the string.
 	for i := 0; i < len(s); i++ {
+		// Check if the character needs encoding. Control characters, slashes,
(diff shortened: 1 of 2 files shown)


json.go
```diff
@@ -6,39 +6,46 @@ const hex = "0123456789abcdef"
...
-			case '\r':
-				dst = append(dst, '\\', 'r')
-			case '\t':
-				dst = append(dst, '\\', 't')
-			default:
-				if b >= 0x20 {
-					dst = append(dst, b)
-				} else {
-					dst = append(dst, '\\', 'u', '0', '0', hex[b>>4], hex[b&0xF])
+	for i := 0; i < len(s); i++ {
```
rs: I feel like you're incrementing `i` too much if the next condition is true here.

rs: Nevermind

rs:
That's a great optim. Could you please add a few comment explaining what you're doing (i.e. when you switch between the two technics) to make it easier to understand?

tidwall:
Thanks for the merge. Nice project btw. :)

rs:
Thanks to you!

PR #7, opened 2017-06-25
JSON encoding further optimization

rs:
Optimize JSON encoding even further

Last optimization was for JSON string with no character to encode. This
version focuses on strings with some chars to encode, trying to apply
the same trick for substrings that do not need encoding.

```
benchmark                old ns/op     new ns/op    delta
.../EncodingLast-8       62.8          61.1         -2.71%
.../RuneFirst-8          164           129          -21.34%
.../RuneMiddle-8         133           96.9         -27.14%
.../RuneLast-8           81.9          73.5         -10.26%
.../NoEncoding-8         60.2          51.3         -14.78%
.../EncodingFirst-8      140           116          -17.14%
.../EncodingMiddle-8     112           86.4         -22.86%
```

cc @tidwall

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,11 +255,11 @@ Some settings can be changed and will by applied to all loggers:
 All operations are allocation free (those numbers *include* JSON encoding):
 
 ```
-BenchmarkLogEmpty-8        100000000	    19.1 ns/op	   0 B/op       0 allocs/op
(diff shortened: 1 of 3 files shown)


tidwall:
Even better! It's working perfectly from my side.

PR #13, opened 2017-08-15
Implement the encoding.TextUnmarshaler interface for log level.

Kirguir:
Implement the encoding.TextUnmarshaler interface for log level from its string representation.

### DIFF
--- a/log.go
+++ b/log.go
@@ -74,6 +74,7 @@ import (
 	"strconv"
 	"sync/atomic"
 
+	"fmt"
 	"github.com/rs/zerolog/internal/json"
 )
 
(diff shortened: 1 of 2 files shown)


rs:
Sorry for the delay, I missed this PR.

What's the use case for this?

rs:
Might be dup of #48

rs:
Fixed by #57

maroux:
Hm, just ran into this as well. `ParseLevel` is a custom method that needs to be called. Implementing `TextUnmarshaler` would let other libraries (e.g. https://github.com/BurntSushi/toml) unmarshal log level automatically based on that stdlib interface.

maroux:
Just opened a PR, we can discuss further there :)

PR #17, opened 2017-10-10
Support numeric log-levels

sean-:
For the sake of compatibility with upstream tools, extend the log level to emit numeric log levels intead of string constants.  Implement as a bitmask flagset.  Expose the flagset as a dedicated type and as a global variable for direct manipulation by callers.


rs:
I don't think zerolog should try to be compatible with all other existing logger implementations. That would make the code very hard to maintain.

Check #25, with this new hook system you might be able to fill the gap without changing zerolog.

rs:
Closing this for now. Reopen if you disagree with my previous comment.

PR #18, opened 2017-10-15
Speed up operations when logging is disabled.

rasky:
Low-level optimizations to help the compiler generate better code
when logging is disabled. Measured improvement is ~30% on amd64
(from 21 ns/op to 16 ns/op).

Unfortunately, this causes a small API breakage: you can't anymore
write `New(os.Stderr).Warn()` (or similar logging function that
returns `*Event`) because the pointer method needs the receiver
to be assigned to a variable. So the code has to be rewritten as 
`log := New(os.Stderr); log.Warn()`. `New(os.Stderr).With()` still
works, it's only functions that return `*Event` that are broken.

### DIFF
--- a/event.go
+++ b/event.go
@@ -21,11 +21,10 @@ var eventPool = &sync.Pool{
 // Event represents a log event. It is instanced by one of the level method of
 // Logger and finalized by the Msg or Msgf method.
 type Event struct {
-	buf     []byte
-	w       LevelWriter
-	level   Level
-	enabled bool
-	done    func(msg string)
+	buf   []byte
+	w     LevelWriter
+	level Level
+	done  func(msg string)
 }
 
 // LogObjectMarshaler provides a strongly-typed and encoding-agnostic interface
@@ -49,12 +48,11 @@ func newEvent(w LevelWriter, level Level, enabled bool) *Event {
 	e.buf[0] = '{'
 	e.w = w
 	e.level = level
-	e.enabled = true
 	return e
 }
 
 func (e *Event) write() (err error) {
-	if !e.enabled {
+	if e == nil {
 		return nil
 	}
 	e.buf = append(e.buf, '}', '\n')
@@ -66,7 +64,7 @@ func (e *Event) write() (err error) {
 // Enabled return false if the *Event is going to be filtered out by
 // log level or sampling.
 func (e *Event) Enabled() bool {
-	return e.enabled
+	return e != nil
 }
 
 // Msg sends the *Event with msg added as the message field if not empty.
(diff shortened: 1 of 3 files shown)


rasky:
These are the benchmarks I used to test the PR:

```go
func BenchmarkZeroLog_On(b *testing.B) {
	var a int16 = 1
	var x string = "2"
	var c bool = false

	zerolog.SetGlobalLevel(zerolog.WarnLevel)
	logger := zerolog.New(ioutil.Discard)

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		logger.Warn().Int16("a", a).Str("x", x).Bool("c", c).
			Msg("logging test")
	}
}

func BenchmarkZeroLog_Off(b *testing.B) {
	var a int16 = 1
	var x string = "2"
	var c bool = false

	zerolog.SetGlobalLevel(zerolog.WarnLevel)
	logger := zerolog.New(ioutil.Discard)

	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		logger.Debug().Int16("a", a).Str("x", x).Bool("c", c).
			Msg("logging test")
	}
}
```

rasky:
@rs ping

rs:
Sorry for the delay. For some reason I missed your initial PR.

rasky:
It generates better code. This is the code generated by a `log.Log()` call after my patch with pointer receivers, using go tip:

```
  log_test.go:135	0x1111f11		488d9c2448010000		LEAQ 0x148(SP), BX
  log_test.go:135	0x1111f19		48891c24			MOVQ BX, 0(SP)
  log_test.go:135	0x1111f1d		e87e8fffff			CALL github.com/rs/zerolog.(*Logger).Log(SB)
  log_test.go:135	0x1111f22		488b442408			MOVQ 0x8(SP), AX
```

This is the one generated with value receivers:

```
  log_test.go:135	0x1111f94		488db42448010000		LEAQ 0x148(SP), SI
  log_test.go:135	0x1111f9c		48896c24f0			MOVQ BP, -0x10(SP)
  log_test.go:135	0x1111fa1		488d6c24f0			LEAQ -0x10(SP), BP
  log_test.go:135	0x1111fa6		e84d20f4ff			CALL 0x1053ff8
  log_test.go:135	0x1111fab		488b6d00			MOVQ 0(BP), BP
  log_test.go:135	0x1111faf		e8ec8effff			CALL github.com/rs/zerolog.Logger.Log(SB)
  log_test.go:135	0x1111fb4		488b442440			MOVQ 0x40(SP), AX
```

The extra call is a call into the runtime to create a copy of the `Logger` object with all the required write barriers, etc. It would be great if the Go compiler was smarter and realized that it doesn't need a copy, but it's not there yet.

Some context: last year, I wrote a library similar to zerolog for its API, but the goal was to completely minimize logging overhead for calls that produce no output because logging is disabled or the log level is higher (I didn't care about performance when some logging is actually done). My library is still much faster than zerolog (~ 4ns/op), but this PR brings it closer. I want to switch to zerolog, but I would like it to perform better when logging is disabled.

Issue #16, opened 2017-09-25
With().Str(...).Logger() fails while With().Timestamp().Str(...).Logger() works


fiatjaf:
If I try to bind some parameters to a context without calling `Timestamp()` on it then it will fail at print time (I'm logging to Stdout with `zerolog.New(os.Stderr).Output(zerolog.ConsoleWriter{Out: os.Stderr})`).

The error is:

```
zerolog: could not write event: invalid character 'f' looking for beginning of object key string
```

`f` is the first character of the first bound parameter (the key). If I change the parameter the "invalid character" also changes.

fiatjaf:
Smallest failing snippet:

```bash
fiatjaf@cantillon ~/c/test> cat x.go 
package main

import (
        "os"

        "github.com/rs/zerolog"
)

func main() {
        log := zerolog.New(os.Stderr).Output(zerolog.ConsoleWriter{Out: os.Stderr})

        bound := log.With().Str("x", "y").Logger()
        bound.Info().Msg("fails")

        bound = log.With().Timestamp().Str("x", "y").Logger()
        bound.Info().Msg("doesn't fail")
}
fiatjaf@cantillon ~/c/test> 
fiatjaf@cantillon ~/c/test> 
fiatjaf@cantillon ~/c/test> go run x.go 
zerolog: could not write event: invalid character 'x' looking for beginning of object key
string2017-09-25T16:53:07Z |INFO| doesn't fail x=y
```

toravir:
Here the fix for this error - will create a pr with this fix shortly..
`
diff --git a/log.go b/log.go
index 83d1927..05a371e 100644
--- a/log.go
+++ b/log.go
@@ -166,7 +166,7 @@ func (l Logger) Output(w io.Writer) Logger {
 func (l Logger) With() Context {
        context := l.context
        l.context = make([]byte, 0, 500)
-       if context != nil {
+       if context != nil && len(context) != 0 {
                l.context = append(l.context, context...)
        } else {
                // first byte of context is presence of timestamp or not
`

toravir:
I am sorry there is typo in this diff - it should be "if l.context != nil {" - NOT l2.context..

rs:
Fixed by 9a65e7ccd2bad9082586a4de21f914648416c90f

PR #20, opened 2017-10-29
Fix for https://github.com/rs/zerolog/issues/16

toravir:
Also includes tests for Output()


toravir:
I am sorry for the confusing 3 commits - for a simple fix.

toravir:
@rs - can you pickup this PR.
Also, what are your thoughts on adding BSON support in addition to the current JSON ? I can continue the proto-type if that is something you think is inline with the goals of this repo.

rs:
I'd like to keep the code as simple as possible. If you think adding BSON support could be done without adding too much complexity I'm open to have a look.

PR #24, opened 2017-11-27
Added hook support

rcoelho:
Hi Olivier, I found zerolog recently and it looks great! Thanks for sharing it.

I wanted to do something it didn't support, then I added hooks.

This makes it possible for the user to implement adding caller information and stack traces, satisfies #9, #11, #22, and keeps the API clean and simple.

This also supports what's being asked for on #12 (example included in this PR, see the readme file).

Let me know if I overlooked anything (such as copying the hook slice over, or resetting it somewhere).
I can provide more tests, benchmarks and update the docs if what's in this PR isn't enough.

Please let me know if this makes sense.

Thanks!

PR #25, opened 2017-11-27
Add hook support

rcoelho:
Hi @rs  I improved my implementation (#24), making it faster and removing all allocations.

Since this adds a change to the Event struct (adds the hasLevel field), I'm sending a separate PR.
Please check (and hopefully merge) #24 before this one.

Thanks!

### DIFF
--- a/event.go
+++ b/event.go
@@ -18,7 +18,7 @@ var eventPool = &sync.Pool{
 	},
 }
 
-type hookRunner func(e *Event, msg string)
+type hookRunner func(e *Event, level Level, msg string)
 
 // Event represents a log event. It is instanced by one of the level method of
 // Logger and finalized by the Msg or Msgf method.
@@ -80,10 +80,10 @@ func (e *Event) Msg(msg string) {
 		return
 	}
 	if len(e.hr) > 0 {
-		e.hr[0](e, msg)
+		e.hr[0](e, e.level, msg)
 		if len(e.hr) > 1 {
 			for _, hook := range e.hr[1:] {
-				hook(e, msg)
+				hook(e, e.level, msg)
 			}
 		}
 	}
@@ -108,10 +108,10 @@ func (e *Event) Msgf(format string, v ...interface{}) {
 	}
 	msg := fmt.Sprintf(format, v...)
 	if len(e.hr) > 0 {
-		e.hr[0](e, msg)
+		e.hr[0](e, e.level, msg)
 		if len(e.hr) > 1 {
 			for _, hook := range e.hr[1:] {
-				hook(e, msg)
+				hook(e, e.level, msg)
 			}
 		}
 	}
--- a/event.go
+++ b/event.go
@@ -23,11 +23,12 @@ type hookRunner func(e *Event, level Level, msg string)
 // Event represents a log event. It is instanced by one of the level method of
 // Logger and finalized by the Msg or Msgf method.
 type Event struct {
-	buf   []byte
-	w     LevelWriter
-	level Level
-	done  func(msg string)
-	hr    []hookRunner
+	buf      []byte
+	w        LevelWriter
+	level    Level
+	done     func(msg string)
+	h        []Hook
+	hasLevel bool
 }
 
 // LogObjectMarshaler provides a strongly-typed and encoding-agnostic interface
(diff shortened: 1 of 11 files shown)


rcoelho:
Some benchmarks:

Before the improvements:

```
BenchmarkHooks/Nop/Single-8             20000000                56.3 ns/op            40 B/op          2 allocs/op
BenchmarkHooks/Nop/Multi-8              20000000                81.8 ns/op            80 B/op          3 allocs/op
BenchmarkHooks/Simple-8                 20000000                91.3 ns/op            40 B/op          2 allocs/op
```

After the first commit:

```
BenchmarkHooks/Nop/Single-8             30000000                35.4 ns/op            16 B/op          1 allocs/op
BenchmarkHooks/Nop/Multi-8              30000000                45.1 ns/op            32 B/op          1 allocs/op
BenchmarkHooks/Simple-8                 30000000                56.6 ns/op            16 B/op          1 allocs/op
```

Aftere the second commit:

```
BenchmarkHooks/Nop/Single-8             50000000                22.4 ns/op             0 B/op          0 allocs/op
BenchmarkHooks/Nop/Multi-8              50000000                25.8 ns/op             0 B/op          0 allocs/op
BenchmarkHooks/Simple-8                 30000000                42.5 ns/op             0 B/op          0 allocs/op
```

event.go
```diff
@@ -21,10 +21,12 @@ var eventPool = &sync.Pool{
...
-	buf   []byte
-	w     LevelWriter
-	level Level
-	done  func(msg string)
+	buf      []byte
+	w        LevelWriter
+	level    Level
+	done     func(msg string)
+	h        []Hook
+	hasLevel bool
```
rs: What about adding a new `NoLevel` level before `DebugLevel` in the enum so we don't need this extra field?

event.go
```diff
@@ -75,6 +78,14 @@ func (e *Event) Msg(msg string) {
...
 		return
 	}
+	if len(e.h) > 0 {
+		e.h[0].Run(e, e.level, e.hasLevel, msg)
+		if len(e.h) > 1 {
+			for _, hook := range e.h[1:] {
+				hook.Run(e, e.level, e.hasLevel, msg)
+			}
+		}
+	}
```
rs: Could you factorize this so you don't repeat the same code in `Msgf`?

log.go
```diff
@@ -155,6 +173,10 @@ func (l Logger) Output(w io.Writer) Logger {
 	l2 := New(w)
 	l2.level = l.level
 	l2.sampler = l.sampler
+	if len(l.hooks) > 0 {
+		l2.hooks = make([]Hook, len(l.hooks), cap(l.hooks))
```
rs: You don't need to set the cap if equal to the size.

log.go
```diff
@@ -155,6 +173,10 @@ func (l Logger) Output(w io.Writer) Logger {
...
+		copy(l2.hooks, l.hooks)
```
rs: `append` should be slightly faster in this context (with len=0 on prev line indeed).

rs:
Do you think we could leverage the `Event.done` property which is already executed in `Msg` and `Msgf` to implement this?

event.go
rcoelho: I thought about doing it, but wanted to change as little as possible. I'll send some more commits your way.

rs: It would let us remove the ugly `addLevelField` param too.

rcoelho: This part is done. The code is simpler now (benchmarks say it is faster, too).
After the merge could you please handle the NoLog case in console.go?
Please take a look at syslog.go, too.

log.go
```diff
@@ -274,9 +308,7 @@ func (l *Logger) WithLevel(level Level) *Event {
 //
 // You must call Msg on the returned event in order to send the event.
 func (l *Logger) Log() *Event {
```
rs: The description if this method is no longer true.

rcoelho: Sorry, I didn't understand this comment. It is still necessary to call Msg on the returned event.
Or are you referencing the comment "We use panic level..."? It's removed now.
Or something else entirely?

rs: Never mind, I read too fast.

rs:
Will you address other comments in a subsequent commit?

rcoelho:
I will, but not today. Thanks for the reviews!

event.go
rcoelho: Event.Msgf

log.go
rcoelho: I'm just appending now.

log.go
rcoelho: Yeah, using append now, and without using make first.

rcoelho:
> Do you think we could leverage the Event.done property which is already executed in Msg and Msgf to implement this?

Possible, but would complicate things. I think that it is good the way it is now.
Ready to merge if all changes look good to you.

PR #26, opened 2017-12-09
Added possibility not to log ts when writing to syslog

shmel1k:
Hi. I use ConsoleWriter for writing to syslog

According to syslog spec 'https://tools.ietf.org/html/rfc5424#section-6.2.3.1' syslog always writes Timestamp as zerolog does.
After all, I get such kind of log messages:
```
myhostname 1512851215:048 2017-12-09T23:26:48.287291+03:00 host daemon_name[32727]: 2017-12-09T23:26:48+03:00 |DEBU| Starting daemon_name
```

When I do not use logger.With().Timestamp(), I get <nil> instead of TS.

This pullrequest allows to disable TS writing to syslog when Syslog flag is set.

### DIFF
--- a/console.go
+++ b/console.go
@@ -35,6 +35,7 @@ var consoleBufPool = sync.Pool{
 type ConsoleWriter struct {
 	Out     io.Writer
 	NoColor bool
+	Syslog  bool
 }
 
 func (w ConsoleWriter) Write(p []byte) (n int, err error) {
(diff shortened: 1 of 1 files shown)


rs:
May I ask why you use ConsoleWriter to write to syslog? ConsoleWriter is an ineffective writer used for human readable logs on console. It should not be used in production.

shmel1k:
Unfortunately, I do not have  any log aggregator to make them human-readable and I have to use human-readable format in my daemon.
So, due to ConsoleWriter inefficiency, the best way is to use json format?

rs:
Yes, and you reformat when you tail them if needed. This writer can be used for that if you want, but not necessarily in producer.

shmel1k:
ok, so i think this pr can be closed.

thnx!

rs:
sure, np.

PR #27, opened 2017-12-14
Add missing WithLevel method to log package

nogoegst:
Other methods are also missing. Should they be here too?


rs:
Thx for your contrib.

Issue #21, opened 2017-11-01
Binary format support


toravir:
[Enhancement]
having a binary logging support will be useful.. meaning keeping json formatted byte array means the code does a strconv for each data type. If we add an option to have binary log support, it will reduce the cpu and memory overhead. I did this simple test:

For Int type - current test passes '0' - it takes 31.2 ns/op

instead, if i put in MaxInt64 as field value, the time is 44ns/op:
BenchmarkLogFieldType/Int-8         	30000000	        44.0 ns/op

I attribute it to strconv() taking longer time to generate all the digits. 

The conversion of the binary log to a readable format can be done offline.

I did a small code snippet to do comparison of JSON/BSON - the eval code and results are available here: https://github.com/toravir/bson-eval

My proposal is that zerolog can be configured either to use JSON or BSON. All the BSON code will be added into internal/bson/ directory.

Please reply if you think this is useful addition to zerolog.

rs:
Might be interesting.

toravir:
@rs - just keeping your posted.. am more than half-way done with binary support - you could see the changes here: https://github.com/toravir/zerolog/commits/binary

I could use review comments...

rs:
I have a few questions:
* Why CBOR instead of BSON? (This is a candid question, I don't know much about CBOR)
* How do you plan to plug the `cbor` package into `event`/`context`?
* Why did you refactor the `json` package? I'm not against it but I don't think it belongs to this PR (and it will make the review harder).
* Is `cbor` package custom made for this PR or copied/adapted from another project? If the later, we need to put copyright info.
* The global binary flag can be confusing. It gives the impression that it will turn all logging into binary, even existing instances. What it really does is changing the default encoding. Its name (and doc) should reflect that.

I did not review the `cbor` package implementation yet. I need some time to read the spec etc. I will save that for when you are ready for the PR.

Good job!

toravir:
@rs:
Answers inline:
> Why CBOR instead of BSON? (This is a candid question, I don't know much about CBOR)
CBOR is more flexible with multiple different data types that BSON - bson is specific to how mongodb sends json objects across the wire. Plus CBOR is a RFC Standard. I had posted on JSON forum (json@yahoogroups.com) and got a recommendation to use CBOR. FYI - CBOR RFC: https://tools.ietf.org/html/rfc7049

> How do you plan to plug the cbor package into event/context?
[...]

PR #37, opened 2018-02-10
Binary format support

toravir:
Submitting the PR for #21 with two minor things in the todo items:
1. Changes to documentation (README.md)
Because after code review, will write the docs.

2. When doing reflection, skip using JSON and write a binary based encoder
Writing an equivalent of json.Marshal() would take more time and careful testing.


rs:
All the "if binary then cbor else json" everywhere is to avoid using an interface and leave a chance for the compiler to inline those calls? It make the code hard to maintain, we should find a better solution.

toravir:
I went back and forth trying to decide about it - if we decide to use the interface, go will have to alloc memory for the arguments - since it is not known where the interface will point to.

I kept the signatures of json and cbor apis very similar so that if we want to do interface, can be done.

I didn't find any other way - but i will post of go forum asking for alternatives.

rs:
Here is another approach: a6bc163a8760e7aee596b51c977c543612d1d708

The runtime cost is lighter than condition and the code is simpler. The downside is that you need to use build tag to enable binary (e.g.: `go build -tag zerolog_binary`).

toravir:
That is one option. My idea was that user will switch ON binary logging when deploying the code in production – aka some sort of a runtime knob.

 

I’ve posted on go forum - https://forum.golangbridge.org/t/two-providers-one-caller/8118 - lets see if there are other ways other than interfaces.

rs:
Most people would have a separate build for production. So the tag approach would work.

toravir:
Sure Olivier, will add the build tag.

toravir:
one final thing i am still debugging - allocs/op for ArrayObj is non-zero ...

toravir:
@rs - the code is ready for merge - fixed the last issue. Here's the benchmark results - highlighting the difference. https://docs.google.com/spreadsheets/d/1SDFzgPKkkF1xckoNh-5RlCH1XRvgrKX_r-co3u_Bupo/edit?usp=sharing

There are two sheets:
1. JSON Logging (base code Vs with these changes)
2. In my forked branch - JSON Logging Vs Binary Logging

Two things i will commit in the next:
1. Documentation
2. Replace the call to json.Marshal() in CBOR for Objects not implementing the marshaller Interface
3. Decoder tool that can take a file containing binary logs and print out json (text)

Thx

log.go
```diff
@@ -363,3 +363,37 @@ func (l *Logger) should(lvl Level) bool {
...
+func binaryFmt(p []byte) bool {
+	if p[0] > 0x7F {
+		return true
+	}
+	return false
+}
+
+//DecodeIfBinaryToString - converts a binary formatted log msg to a
+//JSON formatted String Log message - suitable for printing to Console/Syslog etc
+func DecodeIfBinaryToString(in []byte) string {
```
rs: I would not export those methods.

syslog.go
```diff
@@ -26,11 +28,13 @@ func SyslogLevelWriter(w SyslogWriter) LevelWriter {
 }
 
 func (sw syslogWriter) Write(p []byte) (n int, err error) {
+	p = DecodeIfBinaryToBytes(p, true)
```
rs: I'm not sure it is relevant to do that. What is the point to enable binary logging if the writer silently converts it back to JSON?

json_encoder.go
```diff
@@ -0,0 +1,184 @@
+// +build !enable_binary_log
+
```
rs: I would have named the two encoders `encoder_json.go` and `encoder_cbor.go` so they stay grouped together.

rs:
Do you know why some of the JSON parts are so much worse after the change?

array.go
```diff
@@ -16,7 +14,7 @@ var arrayPool = &sync.Pool{
 }
 
 type Array struct {
-	buf []byte
+	buf      []byte
```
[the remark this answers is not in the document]
toravir: oops.. will run gofmt and update the code.

hook_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !enable_binary_log
```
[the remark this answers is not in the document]
toravir: hooks does work. these tests compare the output with a json string. So duplicated few of the testcases into binary_tests.go and there i convert binary to json and compare with the output.

other thought i had was to clone each _test.go into a _binary_test.go and do the exact same tests.. but then i decided it will be too many _test.go files - so i went with one binary_tests.go with all binary related tests (hooks, ctxt..)

ctx_test.go
```diff
@@ -1,3 +1,5 @@
+// +build !enable_binary_log
```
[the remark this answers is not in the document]
toravir: see above comment for hook_test.go

internal/json/types.go
```diff
@@ -7,6 +7,17 @@ import (
 	"strconv"
 )
 
+func AppendBeginMarker(dst []byte) []byte {
+	return append(dst, '{')
+}
+
+func AppendEndMarker(dst []byte, terminal bool) []byte {
```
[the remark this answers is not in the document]
toravir: good idea. will introduce appendLineBreak..

log.go
toravir: I think zerolog needs to provide decode method apis. Since CBOR package is internal to zerolog, i thought zerolog can provide one.

syslog.go
toravir: This and console are writing to text destinations (syslog/console), outputing binary is not good - so i thought decoding is needed. in case of writing to these text destinations, we cannot delay the decoding of the binary to json. when writing to file or stream etc we can run a decoder.

An alternative for this is users run another process that takes these logs and decodes it and prints to console/syslog...

json_encoder.go
toravir: i can rename the files to encoder_json,  encoder_cbor.

toravir:
will investigate a bit more on the reason for cpu increase

hook_test.go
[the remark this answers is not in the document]
rs: What about wrapping the wanted output into a noop method in json mode that convert the json into cbor in binary mode?

log.go
rs: If we provide a decoder, to be useful it should take a reader and write into a struct instead converting to json. It would be thus better to expose the cbor package.

I don’t think it is the role of a log lib to do that. If there is not cbor lib available to decode cbor, let’s expose the cbor lib as a separate project.

syslog.go
rs: I get it for console. For syslog we should not implement it in binary mode as it does not make sense.

hook_test.go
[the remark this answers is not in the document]
toravir: Done.

array.go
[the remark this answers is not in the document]
toravir: Done

ctx_test.go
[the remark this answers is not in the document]
toravir: done - enabled this test for binary as well.

internal/json/types.go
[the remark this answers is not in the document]
toravir: done

internal/json/types.go
```diff
@@ -276,3 +287,27 @@ func AppendInterface(dst []byte, i interface{}) []byte {
 	}
 	return append(dst, marshaled...)
 }
+
+func AppendObjectData(dst []byte, o []byte) []byte {
+	if o[0] == '{' {
+		o[0] = ','
```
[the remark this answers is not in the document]
toravir: done

log.go
toravir: good point. no longer expose this api.

syslog.go
toravir: excluded syslog and hlog from binary support.

json_encoder.go
toravir: done

toravir:
@rs - i've updated the same google doc with the new performance numbers. There is a little bit of variation of the ns/op numbers between each run (i've noticed variation of +/- 40ns on my machine) for the same test in the same codebase. For the one case "LogArrayObject" case the difference is about 200ns - i did a bunch of trials with various code combinations and found that the overhead of function calls adds up to it and ArrayObject lots of such function calls (as compared to other benchmarks). These functions are not leaf functions, so they are not getting inlined.

The variation in calculation (+/- 40ns) explains the cases where the cpu usage is lower in spite of the extra function call introduced by this code change.

diode/diode_test.go
```diff
@@ -11,17 +12,20 @@ import (
...
 		fmt.Printf("Dropped %d messages\n", missed)
 	}))
-	w := diode.NewWriter(os.Stdout, d, 10*time.Millisecond)
+	buf := bytes.Buffer{}
+	w := diode.NewWriter(&buf, d, 10*time.Millisecond)
 	log := zerolog.New(w)
 	log.Print("test")
 
 	w.Close()
+	fmt.Println(cbor.DecodeIfBinaryToString(buf.Bytes()))
```
rs: Debug left-over?

hlog/hlog.go
```diff
@@ -1,3 +1,5 @@
+// +build !enable_binary_log
```
rs: Why?

diode/diode_test.go
toravir: I have three options with tests:
1. Disable them for binary (since the tests compare the output with text output).
2. Decode if binary and let the comparison with text output continue..
3. Go with Option 1 - but create another file which does the same test for binary

I chose 2nd option.

hlog/hlog.go
toravir: I made a judgement call that http logger will not want binary encoding - just like how syslog and binary log doesn't go together. Do you think otherwise ?

rs: HTTP handlers are just tools to add context to the logs. The format of the output does not matter.

toravir: my bad. i didn't read hlog in detail... Will read thru and make necessary changes in a few days..

diode/diode_test.go
toravir: A cleaner approach occurred to me:
Since examples show up in documentation, i made it just text - so doc looks clean
Same examples, i made sure that there were tests that tested both ascii and binary - so i split diode_test.go to diode_example_test.go (text only) / diode_test.go (text and binary).

hlog/hlog.go
toravir: Enabled hlog_test.go for binary test too. hlog_example_test.go - i left it just text.

rs:
Bench with current master in JSON mode:

```benchstat old new
name                               old time/op    new time/op    delta
LogEmpty-8                           12.5ns ± 6%    16.4ns ±11%  +30.61%  (p=0.000 n=9+10)
Disabled-8                           2.38ns ± 8%    2.89ns ±12%  +21.18%  (p=0.000 n=10+10)
Info-8                               39.5ns ± 5%    53.7ns ± 8%  +35.92%  (p=0.000 n=10+10)
ContextFields-8                      43.1ns ±11%    53.9ns ± 6%  +25.00%  (p=0.000 n=10+10)
ContextAppend-8                      16.8ns ±14%    19.5ns ± 3%  +16.12%  (p=0.000 n=10+10)
LogFields-8                           223ns ± 9%     223ns ±10%     ~     (p=0.795 n=9+10)
LogArrayObject-8                      473ns ± 6%     567ns ± 2%  +19.71%  (p=0.000 n=9+9)
LogFieldType/Err-8                   40.1ns ± 2%    40.0ns ± 5%     ~     (p=0.572 n=8+10)
LogFieldType/Time-8                   155ns ± 3%     133ns ± 5%  -14.24%  (p=0.000 n=10+10)
LogFieldType/Times-8                 1.11µs ± 5%    0.92µs ± 3%  -17.24%  (p=0.000 n=10+9)
LogFieldType/Dur-8                   57.4ns ± 6%    52.8ns ± 3%   -8.00%  (p=0.000 n=8+8)
LogFieldType/Ints-8                  66.0ns ± 7%    60.8ns ± 8%   -7.81%  (p=0.001 n=9+9)
LogFieldType/Bools-8                 43.5ns ± 8%    43.5ns ± 5%     ~     (p=0.926 n=10+10)
LogFieldType/Interface-8              208ns ± 8%     203ns ±15%     ~     (p=0.098 n=8+9)
LogFieldType/Interface(Object)-8      103ns ± 7%     100ns ± 3%   -3.36%  (p=0.024 n=10+10)
LogFieldType/Object-8                 103ns ± 8%      97ns ± 4%   -6.03%  (p=0.002 n=10+9)
LogFieldType/Bool-8                  29.3ns ± 5%    37.2ns ± 4%  +27.05%  (p=0.000 n=9+8)
LogFieldType/Floats-8                 202ns ± 9%     207ns ± 2%   +2.53%  (p=0.020 n=9+7)
LogFieldType/Durs-8                   363ns ± 5%     362ns ± 7%     ~     (p=0.713 n=9+9)
LogFieldType/Interfaces-8             839ns ± 4%     843ns ± 8%     ~     (p=0.920 n=10+9)
LogFieldType/Int-8                   36.1ns ± 7%    38.8ns ±15%   +7.59%  (p=0.041 n=10+10)
LogFieldType/Str-8                   33.1ns ± 6%    37.9ns ± 3%  +14.64%  (p=0.000 n=10+9)
LogFieldType/Strs-8                  63.2ns ±11%    63.3ns ±10%     ~     (p=0.781 n=10+10)
LogFieldType/Errs-8                  52.5ns ±12%    52.8ns ± 2%     ~     (p=0.181 n=10+7)
LogFieldType/Interface(Objects)-8     905ns ± 5%     931ns ±15%     ~     (p=0.239 n=10+10)
LogFieldType/Float-8                 51.0ns ±11%    51.3ns ± 1%     ~     (p=0.433 n=10+9)
ContextFieldType/Bools-8              165ns ±17%     171ns ±17%     ~     (p=0.492 n=10+10)
ContextFieldType/Int-8                148ns ±17%     163ns ±25%     ~     (p=0.072 n=10+10)
ContextFieldType/Timestamp-8          184ns ±14%     183ns ±11%     ~     (p=0.985 n=10+10)

name                               old alloc/op   new alloc/op   delta
LogEmpty-8                            0.00B          0.00B          ~     (all equal)
Disabled-8                            0.00B          0.00B          ~     (all equal)
Info-8                                0.00B          0.00B          ~     (all equal)
ContextFields-8                       0.00B          0.00B          ~     (all equal)
ContextAppend-8                       0.00B          0.00B          ~     (all equal)
LogFields-8                           0.00B          0.00B          ~     (all equal)
LogArrayObject-8                      0.00B          0.00B          ~     (all equal)
LogFieldType/Err-8                    0.00B          0.00B          ~     (all equal)
LogFieldType/Time-8                   0.00B          0.00B          ~     (all equal)
LogFieldType/Times-8                  0.00B          0.00B          ~     (all equal)
LogFieldType/Dur-8                    0.00B          0.00B          ~     (all equal)
LogFieldType/Ints-8                   0.00B          0.00B          ~     (all equal)
LogFieldType/Bools-8                  0.00B          0.00B          ~     (all equal)
LogFieldType/Interface-8               224B ± 0%      224B ± 0%     ~     (all equal)
LogFieldType/Interface(Object)-8      48.0B ± 0%     48.0B ± 0%     ~     (all equal)
LogFieldType/Object-8                 48.0B ± 0%     48.0B ± 0%     ~     (all equal)
LogFieldType/Bool-8                   0.00B          0.00B          ~     (all equal)
LogFieldType/Floats-8                 0.00B          0.00B          ~     (all equal)
LogFieldType/Durs-8                   0.00B          0.00B          ~     (all equal)
LogFieldType/Interfaces-8              640B ± 0%      640B ± 0%     ~     (all equal)
LogFieldType/Int-8                    0.00B          0.00B          ~     (all equal)
LogFieldType/Str-8                    0.00B          0.00B          ~     (all equal)
LogFieldType/Strs-8                   0.00B          0.00B          ~     (all equal)
LogFieldType/Errs-8                   0.00B          0.00B          ~     (all equal)
LogFieldType/Interface(Objects)-8      640B ± 0%      640B ± 0%     ~     (all equal)
LogFieldType/Float-8                  0.00B          0.00B          ~     (all equal)
ContextFieldType/Bools-8               512B ± 0%      512B ± 0%     ~     (all equal)
ContextFieldType/Int-8                 512B ± 0%      512B ± 0%     ~     (all equal)
ContextFieldType/Timestamp-8           528B ± 0%      528B ± 0%     ~     (all equal)

name                               old allocs/op  new allocs/op  delta
LogEmpty-8                             0.00           0.00          ~     (all equal)
Disabled-8                             0.00           0.00          ~     (all equal)
Info-8                                 0.00           0.00          ~     (all equal)
ContextFields-8                        0.00           0.00          ~     (all equal)
ContextAppend-8                        0.00           0.00          ~     (all equal)
LogFields-8                            0.00           0.00          ~     (all equal)
LogArrayObject-8                       0.00           0.00          ~     (all equal)
LogFieldType/Err-8                     0.00           0.00          ~     (all equal)
LogFieldType/Time-8                    0.00           0.00          ~     (all equal)
LogFieldType/Times-8                   0.00           0.00          ~     (all equal)
LogFieldType/Dur-8                     0.00           0.00          ~     (all equal)
LogFieldType/Ints-8                    0.00           0.00          ~     (all equal)
LogFieldType/Bools-8                   0.00           0.00          ~     (all equal)
LogFieldType/Interface-8               2.00 ± 0%      2.00 ± 0%     ~     (all equal)
LogFieldType/Interface(Object)-8       1.00 ± 0%      1.00 ± 0%     ~     (all equal)
LogFieldType/Object-8                  1.00 ± 0%      1.00 ± 0%     ~     (all equal)
LogFieldType/Bool-8                    0.00           0.00          ~     (all equal)
LogFieldType/Floats-8                  0.00           0.00          ~     (all equal)
LogFieldType/Durs-8                    0.00           0.00          ~     (all equal)
LogFieldType/Interfaces-8              4.00 ± 0%      4.00 ± 0%     ~     (all equal)
LogFieldType/Int-8                     0.00           0.00          ~     (all equal)
LogFieldType/Str-8                     0.00           0.00          ~     (all equal)
LogFieldType/Strs-8                    0.00           0.00          ~     (all equal)
LogFieldType/Errs-8                    0.00           0.00          ~     (all equal)
LogFieldType/Interface(Objects)-8      4.00 ± 0%      4.00 ± 0%     ~     (all equal)
LogFieldType/Float-8                   0.00           0.00          ~     (all equal)
ContextFieldType/Bools-8               1.00 ± 0%      1.00 ± 0%     ~     (all equal)
ContextFieldType/Int-8                 1.00 ± 0%      1.00 ± 0%     ~     (all equal)
ContextFieldType/Timestamp-8           2.00 ± 0%      2.00 ± 0%     ~     (all equal)```

We still have some significant changes in surprising places.

**Comment by toravir** [comment]
i am struggling a bit to nail down the cause of the cpu increase. For instance Bools is the same where as Bool is very different. Also i ran it on a linux machine and got a slightly different result:
```
LogEmpty-4                               28.8ns Â±12%    28.9ns Â±12%     ~     (p=0.735 n=11+11)
Disabled-4                               4.96ns Â± 2%    4.99ns Â± 2%     ~     (p=0.357 n=10+11)
Info-4                                    100ns Â± 6%     108ns Â± 7%   +8.71%  (p=0.000 n=10+11)
ContextFields-4                           111ns Â± 3%     118ns Â± 7%   +6.14%  (p=0.000 n=8+11)
ContextAppend-4                          55.6ns Â± 0%    58.4ns Â± 0%   +4.94%  (p=0.000 n=10+10)
LogFields-4                               371ns Â± 2%     370ns Â± 1%     ~     (p=0.381 n=10+10)
LogArrayObject-4                         1.09Âµs Â± 1%    1.28Âµs Â± 2%  +17.75%  (p=0.000 n=10+11)
LogFieldType/Floats-4                     370ns Â± 6%     381ns Â± 2%   +3.00%  (p=0.024 n=10+10)
LogFieldType/Str-4                       55.5ns Â± 8%    63.9ns Â±10%  +15.26%  (p=0.000 n=9+11)
LogFieldType/Strs-4                      97.6ns Â±12%   102.0ns Â± 8%   +4.57%  (p=0.027 n=11+11)
```
I am trying a eliminate parts of the code and trying to figure out what is causing this.. if you do have some inputs, would be appreciated. I've read most of the blogs/write ups about pprof etc - still not able to get narrow it quickly..

**Comment by rs** [comment]
I'm not sure to understand why either.

File: context.go [hunk 11]
```diff
@@ -90,186 +87,183 @@ func (c Context) Hex(key string, val []byte) Context {
...
 }
 
 // Err adds the field "error" with err as a string to the logger context.
 // To customize the key name, change zerolog.ErrorFieldName.
 func (c Context) Err(err error) Context {
-	if err != nil {
-		c.l.context = json.AppendError(json.AppendKey(c.l.context, ErrorFieldName), err)
-	}
-	return c
+	return c.AnErr(ErrorFieldName, err)
```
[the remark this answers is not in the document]
toravir: done

fields.go
```diff
@@ -14,82 +12,82 @@ func appendFields(dst []byte, fields map[string]interface{}) []byte {
...
+			dst = appendFloat32(dst, val)
 		case float64:
-			dst = json.AppendFloat64(dst, val)
+			dst = appendFloat64(dst, val)
 		case time.Time:
-			dst = json.AppendTime(dst, val, TimeFieldFormat)
+			dst = appendTime(dst, val, TimeFieldFormat)
 		case time.Duration:
-			dst = json.AppendDuration(dst, val, DurationFieldUnit, DurationFieldInteger)
+			dst = appendDuration(dst, val, DurationFieldUnit, DurationFieldInteger)
```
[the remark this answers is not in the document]
toravir: done

internal/cbor/README.MD
```diff
@@ -0,0 +1,74 @@
+Reference:
```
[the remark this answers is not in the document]
toravir: done

toravir:
@rs - builds on travis hits a spurious error - which i am not able to recreate on my end - https://travis-ci.org/rs/zerolog/builds/358665413 i tried to on my mac, linux and a docker with go1.10 and ran 50+ times, still cannot reproduce it.. not sure what would be causing this. Sometimes the same test passes on travis (like https://travis-ci.org/rs/zerolog/builds/358670909).. just letting you know..

rs:
Awesome work @toravir, thanks a lot!

rs:
One last request before change. Instead of `enable_binary_log` can we use `binary_log` or `zerolog_binary` instead. Flags are about enabling, so I feel like the enable prefix is redundant.

toravir:
Yes @rs 'binary_log' is clear, simple and neat tag name. Pushed the changes. appreciate all your feedback.

.travis.yml
```diff
@@ -9,4 +9,5 @@ matrix:
   allow_failures:
       - go: "master"
 script:
-  go test -v -race -cpu=1,2,4 -bench . -benchmem ./...
+  - go test -v -race -cpu=1,2,4 -bench . -benchmem ./...
+  - go test -v -tags enable_binary_log -race -cpu=1,2,4 -bench . -benchmem ./...
```
rs: You missed this one.

toravir: oops yeah. fixed it.

rs:
@toravir FYI: ea1184b

PR #38, opened 2018-02-13
Remove allocations while logging an Array of Objects.

rasky:
I've added a benchmark that reports allocations of arrays of objects. This is before my patch:

```
BenchmarkLogArrayObject-4   	 1000000	      1307 ns/op	    1920 B/op	       9 allocs/op
```

After my patch:
```
BenchmarkLogArrayObject-4   	 3000000	       493 ns/op	       0 B/op	       0 allocs/op
```


rs:
Good catch!

PR #39, opened 2018-02-19
Update for Go 1.10

y0ssar1an:
* add Go 1.10 to `.travis.yml`.
* add quotes to Go versions in `.travis.yml`, because unquoted `1.10` is interpreted
as Go `1.1`
* change `.travis.yml` references from `tip` to `master`. `tip` is a legacy reference coming from the days when the Go project used mercurial instead of git
* fix `go vet` issues, since `go test` now runs vet in parallel
  * change `Example_LevelFlag` to `Example`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,12 @@
 language: go
 go:
-- 1.7
-- 1.8
-- 1.9
-- tip
+- "1.7"
+- "1.8"
+- "1.9"
+- "1.10"
+- "master"
 matrix:
   allow_failures:
-      - go: tip
+      - go: "master"
 script:
   go test -v -race -cpu=1,2,4 -bench . -benchmem ./...
--- a/log/log_example_test.go
+++ b/log/log_example_test.go
@@ -102,7 +102,7 @@ func ExampleFatal() {
 
 // This example uses command-line flags to demonstrate various outputs
 // depending on the chosen log level.
-func Example_LevelFlag() {
+func Example() {
 	setup()
 	debug := flag.Bool("debug", false, "sets log level to debug")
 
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -1,11 +1,12 @@
 language: go
 go:
-- 1.7
-- 1.8
-- 1.9
-- tip
+- "1.7"
+- "1.8"
+- "1.9"
+- "1.10"
```
y0ssar1an: These versions have to be quoted now, because the unquoted `1.10` will be interpreted as Go `1.1`.

.travis.yml
```diff
@@ -1,11 +1,12 @@
...
+- "master"
```
y0ssar1an: `tip` is a legacy reference from when the Go project used mercurial instead of git.

log/log_example_test.go
```diff
@@ -100,32 +99,6 @@ func ExampleFatal() {
...
-	log.Info().Msg("This message appears when log level set to Debug or Info")
-
-	if e := log.Debug(); e.Enabled() {
-		// Compute log output only if enabled.
-		value := "bar"
-		e.Str("foo", value).Msg("some debug message")
-	}
-
-	// Output: {"level":"info","time":1199811905,"message":"This message appears when log level set to Debug or Info"}
-}
```
y0ssar1an: I tried removing the underscore to make `go vet` happy, but I couldn't get this to show up in `godoc`. I think this is because there is no `LevelFlag` function in the `log` package.

rs: Just call it `Example`.

rs:
Thx

PR #42, opened 2018-03-15
Add support for hex-encoded bytes

awishformore:
This PR adds support for hex-encoded bytes by using the `Hex(...)` functions instead of the `Bytes(...)` functions. Most of the time, when logging a byte slice, this is what we want.

I did not find a more efficient way to do this. It's a lot of overhead to manually encode each byte slice to a hex string whenever we want to log it in such a way, so I think the PR has merrit, even if it's a bit of a convenience feature.

The fact it's less efficient to use `Hex(...)` than `Bytes(...)` should not matter, as it is optional to use and it would incure the same overhead when manually converting each time.

### DIFF
--- a/event.go
+++ b/event.go
@@ -1,6 +1,7 @@
 package zerolog
 
 import (
+	"encoding/hex"
 	"fmt"
 	"os"
 	"runtime"
@@ -216,6 +217,15 @@ func (e *Event) Bytes(key string, val []byte) *Event {
 		return e
 	}
 	e.buf = json.AppendBytes(json.AppendKey(e.buf, key), val)
+	return e
+}
+
+// Hex adds the field key with val as a hex string to the *Event context.
+func (e *Event) Hex(key string, val []byte) *Event {
+	if e == nil {
+		return e
+	}
+	e.buf = json.AppendString(json.AppendKey(e.buf, key), hex.EncodeToString(val))
 	return e
 }
 
--- a/event.go
+++ b/event.go
@@ -1,7 +1,6 @@
 package zerolog
 
 import (
-	"encoding/hex"
 	"fmt"
 	"os"
 	"runtime"
@@ -225,7 +224,7 @@ func (e *Event) Hex(key string, val []byte) *Event {
 	if e == nil {
 		return e
 	}
-	e.buf = json.AppendString(json.AppendKey(e.buf, key), hex.EncodeToString(val))
+	e.buf = json.AppendHex(json.AppendKey(e.buf, key), val)
 	return e
 }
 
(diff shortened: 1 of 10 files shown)


awishformore:
I also fixed up some tests that were missing the `Bytes(...)` function call.

event.go
```diff
@@ -219,6 +220,15 @@ func (e *Event) Bytes(key string, val []byte) *Event {
 	return e
 }
 
+// Hex adds the field key with val as a hex string to the *Event context.
+func (e *Event) Hex(key string, val []byte) *Event {
+	if e == nil {
+		return e
+	}
+	e.buf = json.AppendString(json.AppendKey(e.buf, key), hex.EncodeToString(val))
```
rs: You could create a `json.AppendHex` that just does:

```go
func AppendHex(buf, val []byte) []byte {
    const hextable = "0123456789abcdef"
    for _, v := range val {
        buf = append(buf, hextable[v>>4], hextable[v&0x0f])
    }
    return buf
}
```

This would avoid the slice heap allocation done by `hex.EncodeToString`.

awishformore: Done. I also added comments for all the `AppendXXX(...)` functions that were missing them to shut up the linter.

awishformore:
The tests show some weird output, I'll try to figure it out.
```
--- FAIL: TestAppendHex (0.00s)
	string_test.go:89: appendHex(0) = "00", want 00
	string_test.go:89: appendHex(f) = "0f", want 0f
	string_test.go:89: appendHex(10) = "10", want 10
	string_test.go:89: appendHex(f0) = "f0�", want f0
	string_test.go:89: appendHex(ff) = "ff�", want ff
FAIL
exit status 1
FAIL	github.com/rs/zerolog/internal/json	0.057s
```

internal/json/string.go
```diff
@@ -123,6 +125,20 @@ func AppendBytes(dst, s []byte) []byte {
...
+// the encoded string to the input byte slice.
+//
+// The operation loops though each byte and encodes it as hex using
+// the hex lookup table.
+func AppendHex(dst, s []byte) []byte {
+	dst = append(dst, '"')
+	for _, v := range s {
+		dst = append(dst, hex[v>>4], hex[v&0x0f])
+	}
+	dst = append(dst, s...)
```
rs: I don't understand this line.

awishformore: Stupid copy-paste mistake. Fixing it up with the tests now.

awishformore:
It should be OK now.

rs:
Thx for your contrib 👍

awishformore:
Thank you for the nice logging library :+1:

PR #48, opened 2018-03-29
Add function to get Level from string

dvelitchkov:
Basically a reverse of Level's String() method. It's useful, for example, if the log level can be supplied to the application as a command-line param. Not super high-value, but might be a nice-to-have. I've found myself implementing this twice now in my own projects.

### DIFF
--- a/log.go
+++ b/log.go
@@ -148,6 +148,27 @@ func (l Level) String() string {
 	return ""
 }
 
+// LvlFromString returns the log level that corresponds to the supplied string value. The values are the same as returned by Level's String() method.
+func LvlFromString(l string) Level {
+	switch l {
+	case "debug":
+		return DebugLevel
+	case "info":
+		return InfoLevel
+	case "warn":
+		return WarnLevel
+	case "error":
+		return ErrorLevel
+	case "fatal":
+		return FatalLevel
+	case "panic":
+		return PanicLevel
+	case "":
+		return NoLevel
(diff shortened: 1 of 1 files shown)


log.go
```diff
@@ -148,6 +148,27 @@ func (l Level) String() string {
...
+	case "error":
+		return ErrorLevel
+	case "fatal":
+		return FatalLevel
+	case "panic":
+		return PanicLevel
+	case "":
+		return NoLevel
+	}
+	return NoLevel
```
rs: Shouldn't this case return an error instead?

dvelitchkov: Assuming you're talking about the `return` statement - sure - I was just mirroring the other API. I can make it return an error instead.

dvelitchkov:
Hmm I'm not sure why one unit test failed in the travis ci build. I can't reproduce the failure locally and it just doesn't make sense that my change would cause that particular test to fail - can you run a rebuild?

log.go
```diff
@@ -148,6 +148,27 @@ func (l Level) String() string {
 	return ""
 }
 
+// LvlFromString returns the log level that corresponds to the supplied string value. The values are the same as returned by Level's String() method.
+func LvlFromString(l string) (Level, error) {
```
rs: Can you call it `ParseLevel` instead?

rs:
Fixed by #57

PR #49, opened 2018-03-30
Support for adding IP Address/Prefix + stream based decoder

toravir:
Summary of Changes:
1. Registered the tags in cbor iana (https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml) and code changes to use the registered tags.
2. Added new apis IPAddr(), IPPrefix(), MACAddr() to add to log msgs.
3. Stream based CBOR decoder - simpler code and logic.


internal/cbor/README.md
```diff
@@ -1,74 +1,51 @@
...
-BenchmarkAppendInt/int-Negative-4               100000000               20.7 ns/op             0 B/op          0 allocs/op
-BenchmarkAppendInt/uint64-4                     50000000                36.7 ns/op             0 B/op          0 allocs/op
-BenchmarkAppendInt/int64-4                      30000000                39.6 ns/op             0 B/op          0 allocs/op
-BenchmarkAppendFloat/Float32-4                  50000000                23.9 ns/op             0 B/op          0 allocs/op
-BenchmarkAppendFloat/Float64-4                  50000000                32.8 ns/op             0 B/op          0 allocs/op
-PASS
-ok      github.com/toravir/zerolog/internal/cbor        34.969s
-sprint @ cbor>
+JSON: 920,000 KB
+CBOR: 550,000 KB
```
rs: Would be interesting to include gzipped JSON to the comparison.

toravir: Do you mean using an inline compression ?? like using a writer wrapped with https://golang.org/pkg/compress/zlib/. I had done some tests with it earlier, can include those results.

rs: Yes

toravir: done, also included the sample code used for measurement.

Issue #61, opened 2018-05-11
Feature proposal: optional debug log sections


vith:
A pattern for logging I found handy in the NodeJS world is [the `util.debuglog()` method](https://nodejs.org/api/util.html#util_util_debuglog_section).

Basically you have named sections of the debug log that you can enable the output for individually instead of turning on a firehose of debugging output from every part of the application at once.

So it could work something like
```go
log.Section("foo").Debug().Int("magic", n).Msg("Made some magic")
```

and the output would be enabled by an environment variable:

```console
$ DEBUG_SECTION=foo ./app
```

or alternatively through a setting on the logger in Go. Basically bypassing the Level check for events that have an explicitly enabled section.

To make it even more powerful it could allow specifying any output level for a given section rather than only being for Debug level events, and/or support wildcards for sections that have a hierarchy of name prefixes.

```console
$ LOG_SECTIONS="foo=debug,bar*=info" ./app
```

rs:
Sounds like a good idea. Do you want to give it a try?

JDiPierro:
:+1:! I was about to open an issue requesting "Trace" as a log level but this would work very well in place of that.

deinspanjer:
In a project I'm writing that currently uses logrus, I have a conditional logger that wraps it and lets the application use tests such as:
```
if log.Enabled(LoggingTagOne, LoggingTagTwo) {
  fields := logrus.Fields{"A": getA(), "B": true}
  log.WithFields(fields).Debug(...)
}
```
Having the if block was important to prevent a bunch of unnecessary allocation, execution if it would be wasted, and it provides a nice ability to selectively log for different parts of the app on the fly.

About to open source the project and happy to point you at it if it might be handy.  I added a todo to look at replacing logrus with zerolog as I'm not extremely happy with it so far.

marco-m:
I believe that this is normally called subsystem logging. I am writing this down to help people googling for it :-)

PR #67, opened 2018-05-17
Add filtering by sections.

ccamel:
Proposal for #61.

As I'm interested in having this feature, I implemented it. It's a minimal implementation that meets the functional aspects mentioned by @vith, that is to say it allows specifying any output level for a given section possibly expressed by a regexp.

### DIFF
--- a/log.go
+++ b/log.go
@@ -103,7 +103,9 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"regexp"
 	"strconv"
+	"strings"
 )
 
 // Level defines log levels.
@@ -178,6 +180,7 @@ func ParseLevel(levelStr string) (Level, error) {
 type Logger struct {
 	w       LevelWriter
 	level   Level
+	section string
 	sampler Sampler
 	context []byte
 	hooks   []Hook
@@ -210,6 +213,7 @@ func Nop() Logger {
 func (l Logger) Output(w io.Writer) Logger {
 	l2 := New(w)
 	l2.level = l.level
+	l2.section = l.section
 	l2.sampler = l.sampler
 	if len(l.hooks) > 0 {
 		l2.hooks = append(l2.hooks, l.hooks...)
@@ -248,6 +252,12 @@ func (l *Logger) UpdateContext(update func(c Context) Context) {
 // Level creates a child logger with the minimum accepted level set to level.
 func (l Logger) Level(lvl Level) Logger {
 	l.level = lvl
+	return l
+}
+
+// Section creates a child logger filtering messages based on the existence of the provided section.
+func (l Logger) Section(section string) Logger {
+	l.section = section
 	return l
 }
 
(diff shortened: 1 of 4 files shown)


log.go
```diff
@@ -251,6 +255,12 @@ func (l Logger) Level(lvl Level) Logger {
 	return l
 }
 
+// Section creates a child logger filtering messages based on the existence of the provided section.
+func (l Logger) Section(section string) Logger {
```
rs: Instead of storing the section, you could test the section right away, and either return the same logger or a disabled logger depending on the result.

log.go
```diff
@@ -394,5 +404,26 @@ func (l *Logger) should(lvl Level) bool {
 	if l.sampler != nil && !samplingDisabled() {
 		return l.sampler.Sample(lvl)
 	}
+
+	if l.section != "" {
+		enabledSections := SectionsProviderFunc()
+
+		for _, entry := range strings.Split(enabledSections, ",") {
+			section := strings.Split(entry, "=")
```
rs: Use a parser instead of splitting. This will create a tone of garbage on each execution.

log.go
```diff
@@ -394,5 +404,26 @@ func (l *Logger) should(lvl Level) bool {
...
+			sectionName := section[0]
+			sectionLevel := NoLevel
+
+			if sectionName != "" {
+				if len(section) > 1 {
+					sectionLevel, _ = ParseLevel(section[1])
+				}
+				if matched, _ := regexp.MatchString(sectionName, l.section); matched {
```
rs: I would prefer to limit the pattern matching capabilities to a simple wildcard and avoid the use of regexp.

log_test.go
```diff
@@ -521,3 +522,136 @@ func TestOutputWithTimestamp(t *testing.T) {
...
+		out := &bytes.Buffer{}
+		os.Setenv(SectionsEnvironmentVariableName, "section-1=badlevel")
+		SectionsProviderFunc = envVariableSectionProvider
+		log := New(out).Section("section-1")
+		log.Debug().Msg("")
+		if got, want := decodeIfBinaryToString(out.Bytes()), `{"level":"debug"}`+"\n"; got != want {
+			t.Errorf("invalid log output:\ngot:  %v\nwant: %v", got, want)
+		}
+	})
+}
```
rs: Please add benchmarks.

log.go
ccamel: Sure. 
But doing so prevents to dynamically change the configuration of sections at running time. I like this feature but I presume the issue of performance is more important ?

ccamel: Ah yes, but implementation has the following issue: you can't call `Section()` inline as it returns a `logger` and not a `*logger` as expected by the sub-functions of the call chain (like `Debug()`). This is the same function  signature than `Level()` (I've worked by imitation :smile: ).

rs: That's easy to fix :)

log.go
ccamel: Don't you mind if I add a new file called `section.go` which gathers all that stuff related to sections? (parsing, matching,...)

rs: No on the contrary. Please do.

pjvds:
@ccamel do you need any help to finished this?

plvir-x:
@rs can I do anything to finish this PR?

rs:
- rebase
- add benchmark
- discuss the name: Is "section" what this should be called? Is it clear for everybody?

Issue #75, opened 2018-06-04
Provide equivalent to std library log SetFlags


joshrendek:
It would be nice to be able to show line output with log messages similar to how the standard library does this:

```	log.SetFlags(log.LstdFlags | log.Lshortfile)```

rs:
Could you be more specific please?

joshrendek:
Added an example to README at https://github.com/rs/zerolog/pull/76 , thanks for the headsup

PR #76, opened 2018-06-04
Update readme at example for Caller()

joshrendek:
https://github.com/rs/zerolog/issues/75

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,7 +296,8 @@ log.Logger = log.With().Str("foo", "bar").Logger()
 ### Add file and line number to log
 
 ```go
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -293,6 +293,15 @@ log.Info().Msg("hello world")
 log.Logger = log.With().Str("foo", "bar").Logger()
 ```
 
+### Add file and line number to log
+
+```go
+log.Info().Caller().Msg("hello world")
```
rs: It's better to add it to the context:

```
log.Logger = log.With().Caller().Logger()
```

PR #78, opened 2018-06-21
Add custom error serialization support and provide sane defaults

DusanKasan:
As per https://github.com/rs/zerolog/issues/9 and to offer a different approach from  https://github.com/rs/zerolog/pull/11 and https://github.com/rs/zerolog/pull/35 this PR introduces custom error serialization with sane defaults without breaking the existing APIs.

This is just a first draft and is missing tests. Also, a bit of code duplication which I feel could be reduced but it serves to get the idea across.

It provides global error marshalling by exposing a `var ErrorMarshalFunc func(error) interface{}` in zerolog package that by default is  a function that returns the passed argument. It should be overriden if you require custom error marshalling.

Then in every function that accept error or array of errors `ErrorMarshalFunc` is called on the error and then the result of it is processed like this:
- if it implements `LogObjectMarshaler`, serialize it as an object
- if it is a string serialize as a string
- if it is an error, serialize as a string with the result of `Error()`
- else serialize it as an interface

The side effect of this change is that the encoders don't need the `AppendError/s` methods anymore, as the errors are serialized directly to other types.

### DIFF
--- a/event.go
+++ b/event.go
@@ -18,6 +18,11 @@ var eventPool = &sync.Pool{
 	},
 }
 
+// ErrorMarshalFunc allows customization of global error marshaling
+var ErrorMarshalFunc = func (err error) interface{} {
+	return err
+}
+
 // Event represents a log event. It is instanced by one of the level method of
 // Logger and finalized by the Msg or Msgf method.
 type Event struct {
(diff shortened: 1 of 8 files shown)


event.go
```diff
@@ -18,6 +18,11 @@ var eventPool = &sync.Pool{
 	},
 }
 
+// ErrorMarshalFunc allows customization of global error marshaling
+var ErrorMarshalFunc = func (err error) interface{} {
```
rs: You could have a nil default and skip the type assertion in the default state. It might improve performance when people not using this feature (to be tested though).

event.go
```diff
@@ -239,39 +244,53 @@ func (e *Event) RawJSON(key string, b []byte) *Event {
...
-// Errs adds the field key with errs as an array of strings to the *Event context.
-// If err is nil, no field is added.
-func (e *Event) Errs(key string, errs []error) *Event {
-	if e == nil {
-		return e
+	arr := Arr()
+	for _, err := range errs {
+		marshaled := ErrorMarshalFunc(err)
+		switch m := marshaled.(type) {
+		case LogObjectMarshaler:
```
rs: Don't you miss the `nil` case here?

DusanKasan: Probably not, my reasoning is that nil will be handled by the default case in which it is serialized using the interface method (which will serialize it as "null").

event.go
DusanKasan: I'll write a benchmark and post results.

DusanKasan: So here's the benchmark https://gist.github.com/DusanKasan/6eb7ef0b4e784bbd26a32df0073ec9cc. 

Results on my machine:
```
BenchmarkCalling-4                      20000000                66.7 ns/op            23 B/op          1 allocs/op
BenchmarkCheckingAndCalling-4           20000000                67.5 ns/op            23 B/op          1 allocs/op
BenchmarkCheckingAndNotCalling-4        20000000                65.8 ns/op            23 B/op          1 allocs/op
```

Nil handling would be a marginal speedup in default cases, however if you change the value from nil to something else you'll pay the price of that if that does the checking (again it's marginal so it shouldn't bother no one).

I don't know how to judge this really. However having a nil value as a default may be helpful also because user can't shoot himself in the foot now. In the original design if you assign nil to `ErrorMarshalFunc` it would panic on invocation. 

Thoughts?

DusanKasan:
@rs what is needed/missing for this to be merged? Would be nice to roll this out asap :)

PR #81, opened 2018-07-02
ctx: store logger in context when missing or different that stored one

### LINKED ISSUE #80: Confusing behavior with context.Context integration
First, thanks for this library; the API is easy to use and the `hlog` package is a nice bonus.

We recently started using `zerolog.Ctx` and `(Logger) WithContext` and were surprised to find that `WithContext` modified the context instead of creating a copy of the context with the modified logger. This means that when we set fields in functions that should only be visible to the functions they call, those fields are instead included on _all_ future log message.

This program demonstrates the issue (it happens more naturally in HTTP handlers):

```go
package main

import (
	"context"
	"github.com/rs/zerolog"
	"os"
)

	child(log.WithContext(ctx))
}

}

func main() {

	process(ctx, 1)

	process(ctx, 2)

}
```
```
# expected output

# actual output
{"level":"info","id":1,"message":"processing 2"}
{"level":"info","id":1,"id":2,"message":"hello from child"}
{"level":"info","id":1,"id":2,"message":"done"}
```

I did some digging in the history and it seems like this was added in 6a6144a10bde51f04fcdb0fbb0c50bca28a025f5 to support the `hlog` functions, which are more efficient when they can modify the existing logger and can reasonably expect to exist at the top of the middleware stack.

However, at some later point, all those functions switched to use `(*Logger) UpdateContext`, which achieves the same thing. It seems like `(Logger) WithContext` is now only used to set the initial logger.

I think this behavior goes against the expectations of `context.Context` (modifications are not visible to functions that were not passed the modified context) and makes it difficult to add new logging fields only for sections of a request handler.

Is it possible that `(Logger) WithContext` could be reverted to always return a copy of the context if the logger is updated? 

If not, do you have recommendations on how to work around this behavior? We're currently going to write our own functions to interact with the context, but it would be great if we could stick with the library functions.

rs:
Current implementation stores a copy of the logger as a pointer and
update its content, which is now unecessary since the introduction of
WithContext.

The new WithContext now takes the pointer and store it in the context if
none is stored already or if the last one stored is a different pointer.
This way it is still possible to update the context of a logger stored
in the context, but it is also possible to store a copy of the logger in
a sub-context.

Fix #80


bluekeyes:
:+1:, this looks like it will fix the issue for me. Thank you for the quick response!

PR #82, opened 2018-07-03
replace ErrorMarshalFunc with AppendErrorFunc

DusanKasan:
The ErrorMarshalFunc change that got merged recently introduced a lot of code bloat/repetition and had one big drawback: to implement custom serialization of errors that did not implement the `LogObjectMarshaler` you had to wrap the error in an implementor `LogObjectMarshaler`.

This can be avoided if the error marshalling function would work on a lower level (with encoder directly). It would also reduce the code bloat that was the result of the fact we needed to work with different APIs (context, event, array...) as we now work only with the encoder.

I thought of this change yesterday, but didn't manage to include it in the previous PR. :)

### DIFF
--- a/encoder.go
+++ b/encoder.go
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-type encoder interface {
+type Encoder interface {
 	AppendArrayDelim(dst []byte) []byte
 	AppendArrayEnd(dst []byte) []byte
 	AppendArrayStart(dst []byte) []byte
--- a/event.go
+++ b/event.go
@@ -18,9 +18,21 @@ var eventPool = &sync.Pool{
 	},
 }
 
-// ErrorMarshalFunc allows customization of global error marshaling
-var ErrorMarshalFunc = func (err error) interface{} {
-	return err
+// AppendErrorFunc allows customization of global error marshaling
+var AppendErrorFunc = func(encoder Encoder, buf []byte, err error) []byte {
+	switch m := err.(type) {
+	case nil:
+		return enc.AppendNil(buf)
+	case LogObjectMarshaler:
+		e := newEvent(nil, 0)
+		e.buf = e.buf[:0]
+		e.appendObject(m)
+		buf = append(buf, e.buf...)
+		eventPool.Put(e)
+		return buf
+	default:
+		return enc.AppendString(buf, m.Error())
+	}
 }
 
 // Event represents a log event. It is instanced by one of the level method of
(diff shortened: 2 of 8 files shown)


DusanKasan:
I was also thinking, if we exposed the Encoder interface and allow setting an encoder to the Logger uponed creation (defaulting to JSON) we could get rid of this error special handling case as everyone could use their own encoder (normal use case would be embedding JSON encoder and overriding some of its methods).

This would be more flexible and more idiomatic Go i think. Thoughts @rs?

rs:
Currently we are not using an interface for the encoder but a build tag to switch the concrete type of the internal encoder. The interface is just here for build time consistency check. Switching to interface would impact performance.

encoder.go
```diff
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-type encoder interface {
+type Encoder interface {
```
rs: Please add comments if this is going to be exposed.

event.go
```diff
@@ -18,9 +18,21 @@ var eventPool = &sync.Pool{
 	},
 }
 
-// ErrorMarshalFunc allows customization of global error marshaling
-var ErrorMarshalFunc = func (err error) interface{} {
-	return err
+// AppendErrorFunc allows customization of global error marshaling
```
rs: Missing period at the end of the comment.

DusanKasan:
Oh okay, didn't know there would be a performance hit. Just to understand better, is this:
- because there is some overhead of calling a function on an interface vs struct
- because when done like this the compiler can basically eliminate the function call by embedding the encoder's methods everywhere they are called

As for this PR, how do you feel about the fact that I'm exposing the Encoder interface here?

rs:
Both as issues, and I guess future versions of Go compiler will be able to inline more cases in the current implementation. Using an interface will block most optimizations.

I'm fine with exposing the encoder in this PR. We need to add documentation though.

DusanKasan:
@rs I added a comment on the Encoder interface and revised the comment on AppendErrorFunc. Should I document this change further? If so how/where.

rs:
The more I'm looking at this new interface, the more I'm wondering if we are not making it a harder and more dangerous API to use. Dealing with the encoder is not easy, and if not done properly, can generate invalid JSON (or CBOR).

DusanKasan:
Yeah, I was also thinking about that. But I don't see any other way how to achieve error serialization (see detailed explanation below) with zero allocations without the encoder access. If there is such a way I'm all up for ditching direct encoder access (for the reasons you mentioned).

Problem: 
- You can't/don't want to implement `LogObjectMarshaler` on the error you are serializing. 
- You want to have all errors serialized with the same structure.

Current solution:
- Use ErrorMarshalFunc to accept error and return a `LogObjectMarshal` implementor that wraps around that error and. This introduces that extra allocation we want to avoid.

DusanKasan:
@rs Best way to solve this would be if we could somehow pass Event to the marshaling function in every case (event/array/context). Would this be doable somehow?

rs:
I guess could introduce a fourth type that wraps the others and route the messages to the right implementation. That's where are supposed to start to beg for generics :)

DusanKasan:
What do you have in mind? I mean, would the facade you are suggesting have both event/context and array methods on it? I would be fine with the solution but I feel we should investigate the performance implications (if any).

rs:
Yes, or we could go with an interface and swallow the perf hit only when there is a customized error type handler.

PR #87, opened 2018-07-24
Allow devs to change the width of the logging level column in consolewriter

davidmccormick:
In the console writer, rather than always truncating the log Level to 4 characters, allow the dev to set their own value, padding or trimming as necessary.  Setting a LevelWidth of 0 leaves it as-is without any padding or trimming.

This adds a little flexibility for people who do not like the present way that these fields are truncated for 'DEBU' and 'ERRO'.

### DIFF
--- a/console.go
+++ b/console.go
@@ -31,8 +31,9 @@ var consoleBufPool = sync.Pool{
 	},
 }
 
-// Define the width of the level column, either trim or pad to width, 0 - don't trim or pad.
-var LevelWidth = 4
+// LevelWidth defines the desired character width of the log level column.
+// Default 0 does not trim or pad (variable width based level text, e.g. "INFO" or "ERROR")
+var LevelWidth = 0
 
 // ConsoleWriter reads a JSON object per write operation and output an
 // optionally colored human readable version on the Out writer.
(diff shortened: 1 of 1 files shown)


console.go
```diff
@@ -31,6 +31,9 @@ var consoleBufPool = sync.Pool{
 	},
 }
 
+// Define the width of the level column, either trim or pad to width, 0 - don't trim or pad.
```
rs: Please start the sentence with "LevelWidth defines…"

console.go
```diff
@@ -31,6 +31,9 @@ var consoleBufPool = sync.Pool{
...
+var LevelWidth = 4
```
rs: I'm fine having a sensible default of 0 here.

console.go
```diff
@@ -55,7 +58,13 @@ func (w ConsoleWriter) Write(p []byte) (n int, err error) {
 		if !w.NoColor {
 			lvlColor = levelColor(l)
 		}
-		level = strings.ToUpper(l)[0:4]
+		level = strings.ToUpper(l)
+		if LevelWidth > 0 {
+			for i := len(level); i < LevelWidth; i++ {
+				level += ` `
+			}
```
rs: What about: 
```go
if padding := LevelWidth - len(level); padding > 0 {
    level += strings.Repeat(" ", padding)
}
```

davidmccormick: Hi many thanks - I have updated the with the use of strings.Repeat but have found I still need a little more logic in order to handle the trimming of the column if the desired width is less than length of the level text.  Seen as we have worked out the padding it seems fastest to assume that if we don't pad then we trim the string to the desired length.  

I still need to treat the LevelWidth 0 as a special case trimming to 0 is the opposite of the desired behaviour of turning off the trimming/padding.

davidmccormick:
Many thanks, updates based on feedback.

Issue #84, opened 2018-07-20
Proposal: A customizable ConsoleWriter output


karmi:
Hello, I've recently used `zerolog` in a project, and I like it very much — many thanks!

One thing I wanted to change was the `ConsoleWriter` output: I wanted to format the output differently, I wanted to highlight field values, not field names, etc. So I wrote a custom implementation, looking like this (adapting the output from `zerolog`'s README):

![Screenshot](https://raw.githubusercontent.com/karmi/consolelog/master/screenshot.png)

I've realized that my needs are probably not unique (see https://github.com/rs/zerolog/issues/34), and that it would be nice to expose a customizable `ConsoleWriter`. Therefore, I've extracted what I did in the project into a library, available at https://github.com/karmi/consolelog, since I didn't want to just blindly send code your way.

My question is: is this something you would consider interesting to have in the `zerolog` library itself? If so, I can move the implementation into a patch and send a pull request.

rs:
Yes, as long as it is backward compatible.

rs:
...and does not add extra dependencies.

karmi:
Hi, understood about dependencies, but can you be more specific about "backwards compatibility" — as far as I understand it, the output type is just `io.Writer`, ie. the contract is in the `Write()` method. Or is there any other concern?

rs:
Just make sure that the current API of the ConsoleWriter still works. Seems like it should not be an issue with the way you pass configuration.

PR #92, opened 2018-07-28
Refactored zerolog.ConsoleWriter to allow customization

karmi:
As outlined in https://github.com/rs/zerolog/issues/84, this patch refactors the `zerolog.ConsoleWriter` in two ways:

* The default formatting has been changed to be more sparse and visually attractive
* The formatting logic has been extracted to separate functions, so it can be changed from the user code

I had to change the design of the [prototype](
https://github.com/karmi/consolelog) in order to maintain the backwards compatibility: the prototype has defined default formatting functions (`formatters`) on the concrete instance of `ConsoleWriter`, but without using an initializer, these would be empty. Therefore, I've added a number of defensive checks into `ConsoleWriter.Write()`, and a `ConsoleWriter.Reset()` method. This is not ideal, but I agree that maintaining backwards compatibility is an important goal, and the behaviour is predictable in my testing.

I've added a benchmark test to make sure I inadverently don't harm the performance: the performance has stayed roughly the same. The tests and examples have been run with the `-race` parameter.

I've updated the screenshot to reflect the new look, and added a comprehensive example to the README.


karmi:
The TravisCI [build](https://travis-ci.org/rs/zerolog/builds/409352179) failed with `could not write event: cannot decode event: invalid character 'Â¿' looking for beginning of value--- FAIL: TestConsoleLogger`, which is surprising, since I can't reproduce it locally — the test runs fine bothin isolation as well as part of the whole suite...

rs:
Please use this screenshot: 
pretty

rs:
This travis issue is meh. Any clue?

karmi:
@rs, unfortunately no clue... :/  I've run both of the commands in `.travis.yml` locally, and it runs just fine. What I wanna try is to run the tests inside a Docker container, to replicate the Travis environment more closely. I'll let you know when I do that.

rs:
Oh ok make sense. It’s due to the second line in the script section of the travis file with the tag to test binary logging.

karmi:
Ah, right! Maybe I forgot to check the output of the second test script properly when running it locally.

karmi:
I've just checked again, and indeed, when I run `go test -v -tags binary_log -race ./...` locally, it does fail without the `decodeIfBinaryToBytes()` call — sorry about that!

README.md
```diff
@@ -252,14 +252,38 @@ sublogger.Info().Msg("hello world")
...
 log.Info().Str("foo", "bar").Msg("Hello world")
 
-// Output: 1494567715 |INFO| Hello world foo=bar
+// Output: 3:04PM INF Hello World foo=bar
+```
+
+To customize the configuration and formatting:
+
+```go
+output := zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: time.RFC3339}.Reset()
```
rs: I'm not found of this API with the need to call `Reset` and `SetFormatter` updating a global config. Why can't we store the formatter settings in the `ConsoleWriter` type?

karmi:
I don't like it much either :) 

I haven't found a better way to keep 100% backwards compatibility with existing _user code_ using `zerolog.ConsoleWriter`. As I said, in the github.com/karmi/consolelog prototype, I've used an "initializer" function to create a new `ConsoleWriter`, which is passed to `zerolog.New()`:

```go
output := consolelog.NewConsoleWriter()
logger := zerolog.New(output).With().Timestamp().Logger()
```

In this way, when a user would customize the `consolelog.ConsoleWriter`, like this:

```go
output = consolelog.NewConsoleWriter(
		func(w *consolelog.ConsoleWriter) {
			w.TimeFormat = time.Stamp
			w.SetFormatter(
				zerolog.LevelFieldName,
				func(i interface{}) string { return strings.ToUpper(fmt.Sprintf("%-5s", i)) })			
		},
	)
```

it would registet the customization only for the particular object — in other words, when you would then create another output with the bare `NewConsoleWriter()` function, it would have the default formatting.

In the prototype, all the formatters, the time format, and the parts order have been indeed stored as properties of `ConsoleWriter`; https://github.com/karmi/consolelog/blob/cf60d6a7b0da4c2bd928702aaaee4aeef0f093e5/consolelog.go#L44-L49.

But since I imagine users have existing code where the only "initialization" they do is something like `log.Output(zerolog.ConsoleWriter{Out: os.Stderr})`, I didn't want to break backwards compatibility here, and force them to update the code to switch the literal initialization to the initializer function...

After a lot of thinking and fiddling, I've realized that I can export a `Reset()` method, which would re-initialize the package variables for any _new_ `ConsoleWriter`, mirroring the `init()` logic.

Maybe there's a better way around this? Or maybe I'm over-thinking the backwards compatibility too much?

(Sorry for a delay with the reply, I'm on a vacation.)

rs:
I'm not sure the initializer function design is indicated here. We don't you just expose the `SetFormatter` function and make sure the default formatters (when not set) are the sensible defaults?

rs:
@karmi: are you still willing to work on this PR?

karmi:
Yes, absolutely — I was offline for most of August.

> [Why] don't you just expose the SetFormatter function and make sure the default formatters (when not set) are the sensible defaults?

You mean as the package level function, or on the `ConsoleWriter` object, as it is now? Again, maybe I'm missing something, but what I wanted to achieve was sensible defaults which would _not_ override any "bare" initialization of another console writer the user might do — which has led me to the `Reset()` method...

I'll have another look on the codebase, maybe another solution will jump out at me.

karmi:
Travis failed on Go 1.9 with:

```
--- FAIL: TestNewWriter (0.00s)
	diode_test.go:29: Diode New Writer Test failed. got:, want:{"level":"debug","message":"test"}
```

I've tried to reproduce locally with Go 1.9.2 and the test didn't fail — I'll amend the commit and force push to trigger another build.

karmi:
I've took another stab at refactoring the functionality, and would like to get some feedback on the latest version of [`console.go`](https://github.com/karmi/zerolog/blob/318753efd254c68779c937f29cd56cb4876e8d3f/console.go).

The most important part is removing the global registry (`map`) of formatters, and therefore removing the need for any `Reset()` method, the `init()` function and other weird semantics.

The default formatters are added as package level variables (`consoleDefaultFormatTimestamp`, ...) and custom Formatters are added as regular fields of the `ConsoleWriter` struct. (Example of setting the fields directly [here](https://github.com/karmi/zerolog/blob/318753efd254c68779c937f29cd56cb4876e8d3f/console_test.go#L22-L31), example of using the constructor function option [here](https://github.com/karmi/zerolog/blob/318753efd254c68779c937f29cd56cb4876e8d3f/console_test.go#L42-L57).)

I've plugged this version into a real project where I'm using `zerolog`, and it behaves well. It still doesn't provide a way how to customize formatting a specific field (`component`) in my case, but that's not something which has to be solved right away...

I can imagine there's yet another approach possible, eg. adding a `ConsoleFormatter` struct, and attaching methods to it, and opening up the logic even more, but again, not sure if it has to be solved at once.

README.md
rs: You forgot to update the doc to remove the `.Reset()` here.

README.md
```diff
@@ -8,7 +8,7 @@ Zerolog's API is designed to provide both a great developer experience and stunn
 
 Uber's [zap](https://godoc.org/go.uber.org/zap) library pioneered this approach. Zerolog is taking this concept to the next level with a simpler to use API and even better performance.
 
-To keep the code base and the API simple, zerolog focuses on efficient structured logging only. Pretty logging on the console is made possible using the provided (but inefficient) `zerolog.ConsoleWriter`.
+To keep the code base and the API simple, zerolog focuses on efficient structured logging only. Pretty logging on the console is made possible using the provided (but inefficient) [`zerolog.ConsoleWriter`](#pretty-logging).
 
 ![Pretty Logging Image](pretty.png)
```
rs: This image should be updated with the updated default look of the console writer.

rs:
Looks good. Please fix the doc and I'll merge.

README.md
karmi: Argh, sorry about that, fixed in fc34c71!

karmi:
Thanks! Do you want me to rebase & squash the commits or do you prefer to do it on your side?

wvh:
Hello,

The new `ConsoleWriter` creates `consoleDefaultPartsOrder` on init instead of when creating an instance of itself, so if someone changes the default fields – such as `MessageFieldName` – `ConsoleWriter` will try to print the wrong field names.

For instance, if I set `MessageFieldName` to `"msg"`, the console writer will try to print the original `"message"` field:

```go
var out io.Writer = os.Stdout
zerolog.MessageFieldName = "msg"
out = zerolog.ConsoleWriter{Out: out}
logger = zerolog.New(out).With().Timestamp().Logger()
logger.Info().Msg("this message won't show")
```
results in (more or less):

```
15:26:27.818175 INF %!s(<nil>) at=main.go:171 component=main
```
... note the `nil` message field.

karmi:
@wvh, argh, that's bad... I'll look into it!

karmi:
I did the simplest thing which came to my mind, and made `consoleDefaultPartsOrder` a function, instead of a static slice: https://github.com/rs/zerolog/compare/master...karmi:fix_console_custom_parts

In this way, it's called from `Write()` and therefore picks up the custom name — though I don't like it from the performance point of view much...

What do you think, @rs?

karmi:
@rs, right, that's what I was thinking as well, just wasn't sure. I'll send the patch as a PR.

wvh:
I guess you could also force people to use the New constructor and set it there, but you'd have to make sure all struct fields can be set from the constructor or have setters exposed if you'd make the struct private.

By the way – thanks, I like the new look.

karmi:
@wvh , right, that would solve the problem of initialization, but would break a lot of existing code, and that's what I wanted to prevent at all cost with the refactor...

IngmarStein:
Previously, `ConsoleWriter` could be used concurrently. Now, these package-level configuration options have introduced a data race in `Write` (see below). Is it intended that callers need to add synchronisation around `ConsoleWriter`?
```
==================
WARNING: DATA RACE
Read at 0x00000310ea0d by goroutine 27:
  …/vendor/github.com/rs/zerolog.ConsoleWriter.Write()
      /go/src/…/vendor/github.com/rs/zerolog/console.go:111 +0x656
  …/vendor/github.com/rs/zerolog.(*ConsoleWriter).Write()
      <autogenerated>:1 +0xf0
  …/vendor/github.com/rs/zerolog.levelWriterAdapter.WriteLevel()
      /go/src/…/vendor/github.com/rs/zerolog/writer.go:20 +0x66
  …/vendor/github.com/rs/zerolog.(*levelWriterAdapter).WriteLevel()
      <autogenerated>:1 +0xa4
  …/vendor/github.com/rs/zerolog.(*Event).write()
      /go/src/…/vendor/github.com/rs/zerolog/event.go:80 +0x2e0
  …/vendor/github.com/rs/zerolog.(*Event).msg()
      /go/src/…/vendor/github.com/rs/zerolog/event.go:139 +0x296
  …/vendor/github.com/rs/zerolog.(*Event).Msg()
      /go/src/…/vendor/github.com/rs/zerolog/event.go:110 +0x51
  …

Previous write at 0x00000310ea0d by goroutine 28:
  …/vendor/github.com/rs/zerolog.ConsoleWriter.Write()
      /go/src/…/vendor/github.com/rs/zerolog/console.go:112 +0x684
  …/vendor/github.com/rs/zerolog.(*ConsoleWriter).Write()
      <autogenerated>:1 +0xf0
  …/vendor/github.com/rs/zerolog.levelWriterAdapter.WriteLevel()
      /go/src/…/vendor/github.com/rs/zerolog/writer.go:20 +0x66
  …/vendor/github.com/rs/zerolog.(*levelWriterAdapter).WriteLevel()
      <autogenerated>:1 +0xa4
  …/vendor/github.com/rs/zerolog.(*Event).write()
      /go/src/…/vendor/github.com/rs/zerolog/event.go:80 +0x2e0
  …/vendor/github.com/rs/zerolog.(*Event).msg()
      /go/src/…/vendor/github.com/rs/zerolog/event.go:139 +0x296
  …/vendor/github.com/rs/zerolog.(*Event).Msg()
      /go/src/…/vendor/github.com/rs/zerolog/event.go:110 +0x51
  …
```

karmi:
Hi @IngmarStein , I've tried locally, and was able to reproduce the race when eg. setting the timestamp option, coming from eg. this line: https://github.com/rs/zerolog/blob/8747b7b3a51b5d08ee7ac50eaf4869edaf9f714a/console.go#L103

I'll look into it, it definitely shouldn't require any synchronization from the calling code.

PR #99, opened 2018-08-31
Add the concept of per-category log level

arl:


rs:
Does not look like a PR that was meant to be upstreamed. If yes, please reoopen, add a description explaining what you are trying to solve, and remove s/rs/arl/.

arl:
That's correct, that was not meant to be upstreamed and was opened on the
wrong remote. Sorry for the noise

xmlking:
@arl ability to configure  per-category log level  will be useful for many use-cases ( in case if I want to enable debug level logs only for a specific module in prod) 
Wonder if you contribute back  to zerolog
Thanks

arl:
@xmlking yes that's the idea: being able to assign a category to log events and manage levels on a per-category basis.

The implementation I have aims at having a close-to-zero runtime cost, but it adds a 1-byte field to the Event `struct` to store the category, allowing up to 256 categories, which should be more than enough.

Most of the code is in https://github.com/arl/zerolog/commit/0de11d925f64a50ba7ac70bb364f2d71d0adcc2e I can clean that up a bit and open a PR if that's of interest to the maintainers.

xmlking:
thanks a lot @arl . looking for exactly that to control logging per category with config.yaml or environmental variables, in production

PR #100, opened 2018-09-07
ConsoleWriter: Use GlobalLevel() as a default level

VojtechVitek:
Given
```go
log.Logger = zerolog.New(zerolog.ConsoleWriter{Out: os.Stdout})
zerolog.SetGlobalLevel(zerolog.InfoLevel)

stdlog.SetFlags(0)
stdlog.SetOutput(log.Logger)
stdlog.SetPrefix("")
```

The ConsoleWriter will write "INFO" as a default level for stdlog messages:
```diff
- |????| some log message
+ |INFO| some log message
```

### DIFF
--- a/console.go
+++ b/console.go
@@ -54,18 +54,20 @@ func (w ConsoleWriter) Write(p []byte) (n int, err error) {
 	buf := consoleBufPool.Get().(*bytes.Buffer)
 	defer consoleBufPool.Put(buf)
 	lvlColor := cReset
-	level := "????"
-	if l, ok := event[LevelFieldName].(string); ok {
-		if !w.NoColor {
-			lvlColor = levelColor(l)
-		}
-		level = strings.ToUpper(l)
-		if LevelWidth > 0 {
-			if padding := LevelWidth - len(level); padding > 0 {
-				level += strings.Repeat(" ", padding)
-			} else {
-				level = level[0:LevelWidth]
-			}
+
+	level, ok := event[LevelFieldName].(string)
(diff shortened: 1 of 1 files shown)


rs:
When zerolog is used as a `io.Writer` like so, messsges are logged with no level. You can add a level in the context if want.

VojtechVitek:
@rs I wonder if we should interpret "no level" as INFO by default? Just for the sake of the further consumers of the logs.

Or do you prefer `???`?

rs:
If there is no level in the log message it would be confusing to show a level in the console writer.

VojtechVitek:
Let's close this PR then. Thanks for your feedback.

rs:
You're welcome.

NathanBaulch:
Two workarounds FWIW:

1. Set the event level at the last minute with a hook.
```go
log.Logger.Hook(zerolog.HookFunc(func(e *zerolog.Event, lvl zerolog.Level, msg string) {
	if lvl == zerolog.NoLevel {
		e.Str(zerolog.LevelFieldName, zerolog.InfoLevel.String())
	}
}))
```

2. Wrap the logger and replace the `Write` function.
```go
stdlog.SetOutput(wrapper{log.Logger})

type wrapper struct {
	zerolog.Logger
}

func (l wrapper) Write(p []byte) (n int, err error) {
	n = len(p)
	if n > 0 && p[n-1] == '\n' {
		p = p[:n-1]
	}
	l.Info().Msg(string(p))
	return
}

```

PR #102, opened 2018-09-08
Increased CallerSkipFrameCount in Msgf

manthansharma:
`Msgf` function call `Msg` which add one more hop. CallerSkipFrameCount needs to increase by 1,  to handle that hop.


rs:
This would increase a global var on every call to Msgf. This can’t work.

manthansharma:
I just saw, this issue has already been resolved with #89. Closing PR.

PR #110, opened 2018-09-27
Don't call panic() and os.Exit(1) on .WithLevel()

VojtechVitek:
## Rationale:

I'd like to be able to set `PANIC` or `FATAL` levels explicitly via `logger.WithLevel()`, without having to exit my program.

## Current state:
`logger.Fatal()` calls `os.Exit(1)` .. that's obvious and expected
`logger.Panic()` calls `panic()` .. that's obvious and expected

`logger.WithLevel(zerolog.FatalLevel)` calls `os.Exit(1)` too, which feels wrong
`logger.WithLevel(zerolog.PanicLevel)`  calls `panic()` too, which feels wrong

## Proposal:
I'd like to be able to set an explicit log level via `logger.WithLevel(lvl)` without program flow side effects.

### DIFF
--- a/log.go
+++ b/log.go
@@ -321,9 +321,9 @@ func (l *Logger) WithLevel(level Level) *Event {
 	case ErrorLevel:
 		return l.Error()
 	case FatalLevel:
-		return l.Fatal()
+		return l.newEvent(FatalLevel, nil)
 	case PanicLevel:
-		return l.Panic()
+		return l.newEvent(PanicLevel, nil)
 	case NoLevel:
 		return l.Log()
 	case Disabled:
--- a/log.go
+++ b/log.go
@@ -292,22 +292,24 @@ func (l *Logger) Error() *Event {
 }
 
 // Fatal starts a new message with fatal level. The os.Exit(1) function
-// is called by the Msg method.
+// is called by the Msg method, which terminates the program immediately.
 //
 // You must call Msg on the returned event in order to send the event.
 func (l *Logger) Fatal() *Event {
 	return l.newEvent(FatalLevel, func(msg string) { os.Exit(1) })
 }
 
-// Panic starts a new message with panic level. The message is also sent
-// to the panic function.
+// Panic starts a new message with panic level. The panic() function
+// is called by the Msg method, which stops the ordinary flow of a goroutine.
 //
 // You must call Msg on the returned event in order to send the event.
 func (l *Logger) Panic() *Event {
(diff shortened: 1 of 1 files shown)


rs:
I can't reproduce. The call to `WithLevel` with those levels should not exit or panic until you call `Msg` on the returned event.

VojtechVitek:
@rs right, let me correct myself:

```diff
- logger.WithLevel(zerolog.FatalLevel) calls os.Exit(1) too, which feels wrong
- logger.WithLevel(zerolog.PanicLevel) calls panic() too, which feels wrong
+ logger.WithLevel(zerolog.FatalLevel).Msg("msg") calls os.Exit(1) too, which feels wrong
+ logger.WithLevel(zerolog.PanicLevel).Msg("msg") calls panic() too, which feels wrong
```

I'm proposing to change the above behaviour.

Example:
Let's say I'm writing a "recoverer" middleware where I catch panics from the underlying HTTP handlers. I just want log a message with "PANIC" level to my logger if I recover from panic. But how do I do that without triggering `panic()` again? Do I need to wrap this logic in yet-another defer-recover wrapper?

So I'm proposing 
```go
logger.Panic().Msg("ouch") // calls panic("ouch")
logger.WithLevel(zerolog.Panic).Msg("ouch") // doesn't call panic(), it just logs "ouch" with PANIC level

logger.Fatal().Msg("oops") // prints "oops" to stderr and calls os.Exit(1)
logger.WithLevel(zerolog.Fatal).Msg("oops") // doesn't exit the program, it just logs "oops" with FATAL level
```

rs:
Ok I understand now. I think it's sound. Could you add some comment to the `WithLevel` method to explain this behavior please?

VojtechVitek:
@rs done. I reused stdlib's terminology of os.Exit() and panic().

log.go
```diff
@@ -292,22 +292,24 @@ func (l *Logger) Error() *Event {
...
 //
 // You must call Msg on the returned event in order to send the event.
 func (l *Logger) Panic() *Event {
 	return l.newEvent(PanicLevel, func(msg string) { panic(msg) })
 }
 
-// WithLevel starts a new message with level.
+// WithLevel starts a new message with level. Unlike Fatal and Panic
+// methods, WithLevel does not terminate the program or stop the ordinary
+// flow of a gourotine.
```
rs: I would add something like:

... when used with their respective levels.

VojtechVitek:
@rs `when used with their respective levels` added.

PR #117, opened 2018-11-22
fix typo

shawnps:


hlog/hlog_example_test.go
```diff
@@ -63,7 +63,7 @@ func Example_handler() {
 		hlog.FromRequest(r).Info().
 			Str("user", "current user").
 			Str("status", "ok").
-			Msg("Something happend")
+			Msg("Something happened")
```
marco-m: @shawnps I think that the CI tests are failing because also line 72 needs to be updated; this is a Go testable example (https://blog.golang.org/examples)

PR #118, opened 2018-12-04
Fix race condition in ConsoleWriter

karmi:
When the ConsoleWriter has been customized, the operations accessing
variables in the `Write()` method had race condition, as reported in
https://github.com/rs/zerolog/pull/92#issuecomment-444038726.

A simple mutex lock has been added around the section.

Example code:

```go
package main

import (
	"fmt"
	"os"
	"sync"
	"time"

	"github.com/rs/zerolog"
	"github.com/rs/zerolog/log"
)

func main() {
	var wg sync.WaitGroup

	log.Logger = log.With().Timestamp().Logger().Output(zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: time.RFC3339})

	// output := zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: time.RFC3339}
	// log := zerolog.New(output).With().Timestamp().Logger()

	for i := 0; i < 10000; i++ {
		wg.Add(1)

		go func(i int) {
			defer wg.Done()
			log.Info().Str("foo", "bar").Msg(fmt.Sprintf("Hello from goroutine %d", i))
		}(i)
	}

	wg.Wait()
}
```

### DIFF
--- a/console.go
+++ b/console.go
@@ -36,6 +36,7 @@ var (
 		},
 	}
 
+	mu                       sync.Mutex
 	consoleDefaultTimeFormat = time.Kitchen
 	consoleDefaultFormatter  = func(i interface{}) string { return fmt.Sprintf("%s", i) }
 	consoleDefaultPartsOrder = func() []string {
@@ -96,6 +97,7 @@ func NewConsoleWriter(options ...func(w *ConsoleWriter)) ConsoleWriter {
 
 // Write transforms the JSON input with formatters and appends to w.Out.
 func (w ConsoleWriter) Write(p []byte) (n int, err error) {
+	mu.Lock()
 	if w.PartsOrder == nil {
 		w.PartsOrder = consoleDefaultPartsOrder()
 	}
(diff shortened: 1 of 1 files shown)


karmi:
I guess so. But the assumption is that the console writer is used in a development setting, ie. the performance isn't super-critical...

IngmarStein:
True. I was just thinking if a `sync.Once` may help, but then you couldn't change the settings in between two `Write` calls. A proper solution would be to not store the parameters at the package level but to pass them to the formatters, but I saw you paid attention to keeping backwards compatibility.

By the way, the `consoleDefaultFormatter` variable appears to be unused.

karmi:
I had the same idea, actually, but came to the same conclusion :)

(And yes, the trouble is that I wanted to keep backwards compatibility with the customizable version, so I couldn't rely eg. on a constructor function...)

console.go
```diff
@@ -96,6 +97,7 @@ func NewConsoleWriter(options ...func(w *ConsoleWriter)) ConsoleWriter {
 
 // Write transforms the JSON input with formatters and appends to w.Out.
 func (w ConsoleWriter) Write(p []byte) (n int, err error) {
+	mu.Lock()
```
rs: This lock is ineffective as there is no way for the potential writer of those props to acquire it.

IngmarStein:
I've sketched a lock-free alternative in #120. Please let me know what you think about that approach.

karmi:
I vote for #120, that's a much better solution.

Issue #125, opened 2018-12-19
ConsoleWriter has no colored field name output by default


leeola:
Not a big deal as it's easily fixable with a custom field formatter, but I've noticed that since updating to latest my console writer no longer has colorized output for the field names. Other colors, such as the log level are properly colored.

Version: `github.com/rs/zerolog v1.11.0 h1:DRuq/S+4k52uJzBQciUcofXx45GrMC6yrEbb/CoK6+M=`

rs:
cc @IngmarStein

IngmarStein:
Oops, there's no unit test that covers writing colorized fields.
This one is successful:
```
	t.Run("Write colorized fields", func(t *testing.T) {
		buf := &bytes.Buffer{}
		w := zerolog.ConsoleWriter{Out: buf, NoColor: false}

		d := time.Unix(0, 0).UTC().Format(time.RFC3339)
		_, err := w.Write([]byte(`{"time" : "` + d + `", "level" : "debug", "message" : "Foobar", "foo" : "bar"}`))
		if err != nil {
			t.Errorf("Unexpected error when writing output: %s", err)
		}

		expectedOutput := "\x1b[2m12:00AM\x1b[0m \x1b[33mDBG\x1b[0m Foobar \x1b[2mfoo=\x1b[0mbar\n"
		actualOutput := buf.String()
		if actualOutput != expectedOutput {
			t.Errorf("Unexpected output %q, want: %q", actualOutput, expectedOutput)
		}
	})
```
Could you please provide an example that illustrates the issue?

kevinmcconnell:
I think this is because the console writer now uses `colorFaint` for fields and timestamps, which doesn't seem to have any effect in some terminals.  (At least, it does nothing for me in Terminal.app on macOS -- the `colorFaint` text looks exactly the same as un-colorized text).

Previously it was using cyan & dark grey for those, which both showed up fine.

Would you be interested in it switching back to the old colors?  I'd be happy to submit a PR for that if so.

rs:
Sure, please do.

PR #131, opened 2019-02-07
Don't use faint text in colorized console output

kevinmcconnell:
This is from discussion in #125.

The faint text style doesn't seem to be supported by all terminals, which means the default console output loses most of its coloring on them.

This commit changes the color scheme to more widely-supported colors. This also matches how it looked in earlier versions of this library.

### DIFF
--- a/console.go
+++ b/console.go
@@ -14,11 +14,6 @@ import (
 )
 
 const (
-	colorBold = iota + 1
-	colorFaint
-)
-
-const (
 	colorBlack = iota + 30
 	colorRed
 	colorGreen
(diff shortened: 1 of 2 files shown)


console.go
```diff
@@ -27,6 +22,9 @@ const (
 	colorMagenta
 	colorCyan
 	colorWhite
+
+	colorBold     = 1
+	colorDarkGray = 90
```
kevinmcconnell: I've moved these into the same `const` block because the only thing they have in common with each other is that they don't fit tidily into the iota sequence.  We could pull them back out if that's preferred though.

PR #136, opened 2019-02-25
Allow using custom level field format

soloman1124:
hi, @rs , We are adapting syslog's level keywords (https://en.wikipedia.org/wiki/Syslog) for our level field, and hence it would be really useful if we could have a way to override or customize the level field format.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -13,6 +13,11 @@ var (
 	// LevelFieldName is the field name used for the level field.
 	LevelFieldName = "level"
 
+	// LevelFieldMarshalFunc allows customization of global level field marshaling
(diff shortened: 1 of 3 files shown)


mikeyrcamp:
@soloman1124 I'm not a maintainer, but you may want to add some tests for your change. I know the existing tests get you full coverage, but it doesn't hurt to let the tests server as example especially for your given use case.

rs:
Looks good. Some tests would be welcome yes.

soloman1124:
@rs and @mikeyrcamp thanks for the suggestion, tests added.

PR #138, opened 2019-02-28
MarshalIndent option for ConsoleWriter

mozey:
Reading nested values like stack traces all on one line is hard. 

This PR adds an option to make `ConsoleWriter` use `json.MarshalIndent`


mozey:
Why does the travis check fail? I've ran the tests myself for Go:1.11 with no issues

PR #141, opened 2019-03-08
Fix misleading text about duplicate keys in JSON

### LINKED ISSUE #140: Misleading text about duplicate keys in JSON
[Zerolog documentation](https://godoc.org/github.com/rs/zerolog#hdr-Caveats) says:

> JSON accepts dup keys, the last one prevails

But the [JSON standard](https://tools.ietf.org/html/rfc8259#section-4) says:

> The names within an object SHOULD be unique.
   [...] When the names within an object are not
   unique, the behavior of software that receives such an object is
   unpredictable.  Many implementations report the last name/value pair
   only.  Other implementations report an error or fail to parse the 
   object, and some implementations report all of the name/value pairs,
   including duplicates.

Probably this is not an issue in practice, but then the docs should be rephrased along the lines of “Software accepting JSON logs will typically honor the last one”.

vfaronov:
Per RFC 8259 Section 4, behavior on duplicate keys is unspecified.

Fixes #140.

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,7 +578,7 @@ Log a static string, without any context or `printf`-style templating:
 
 ## Caveats
 
-Note that zerolog does de-duplication fields. Using the same key multiple times creates multiple keys in final JSON:
+Note that zerolog does no de-duplication of fields. Using the same key multiple times creates multiple keys in final JSON:
 
 ```go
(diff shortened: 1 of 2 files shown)


marco-m:
hello @rs ping :-)

PR #144, opened 2019-03-21
Added multistack logging -- wip

mec07:
There is an issue with the way the MarshalStack is implemented right now. It only prints info on the last stack trace -- which is really unhelpful if you have multiple stacks (i.e. you really want to know where the original stack trace came from).

This pull request adds the ability to log all of the stack traces collected by the errors. It is pretty much there, but I've marked as wip because the exact formatting needs to be determined -- probably better to use an inbuilt encoding mechanism of zerolog (instead of json), but I'm not too familiar with the encoding of zerolog... @rs do you have any suggestions? Thanks!

### DIFF
--- a/pkgerrors/stacktrace_test.go
+++ b/pkgerrors/stacktrace_test.go
@@ -27,6 +27,82 @@ func TestLogStack(t *testing.T) {
 	}
 }
 
+func TestLogMultiStack(t *testing.T) {
+	zerolog.ErrorStackMarshaler = MarshalMultiStack
+
+	out := &bytes.Buffer{}
+	log := zerolog.New(out)
+
+	err := errors.Wrap(errors.New("error message"), "from error")
+	log.Log().Stack().Err(err).Msg("")
+
+	got := out.String()
+	want := `\{"stack":"\[\{\"stacktrace\":\[\{\"source\":\"stacktrace_test.go\",\"line\":\"36\",\"func\":\"TestLogMultiStack\"\},.*\{\"stacktrace\".*\],"error":"from error: error message"\}`
+	if ok, _ := regexp.MatchString(want, got); !ok {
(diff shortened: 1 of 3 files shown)


pkgerrors/stacktrace_test.go
```diff
@@ -27,6 +27,82 @@ func TestLogStack(t *testing.T) {
...
+	zerolog.ErrorStackMarshaler = MarshalMultiStack
+
+	out := &bytes.Buffer{}
+	log := zerolog.New(out)
+
+	err := errors.Wrap(errors.New("error message"), "from error")
+	log.Log().Stack().Err(err).Msg("")
+
+	got := out.String()
+	want := `\{"stack":"\[\{\"stacktrace\":\[\{\"source\":\"stacktrace_test.go\",\"line\":\"36\",\"func\":\"TestLogMultiStack\"\},.*\{\"stacktrace\".*\],"error":"from error: error message"\}`
```
mec07: This test isn't passing right now because the json.Marshal is returning a string and I haven't quite cracked how to make the regexp match (got & want look like they should match)... The best thing to do though is to use the proper zerolog encoding rather than using json.Marshal.

mec07: Fixed

mec07:
Actually I was being really dumb, I've worked out how to do the encoding properly now (i.e. without json.Marshal).

mec07:
@rs I would love to hear your thoughts on this. I am using `MarshalMultiStack` from my own fork for now and it has fixed the problem of the incomplete stacktrace from `MarshalStack`.

mec07:
@rs have you had a chance to look at this? Thanks! :)

mec07:
@rs I'd love to hear your thoughts on this.

PR #145, opened 2019-03-27
enable adding raw JSON to an array

nelz9999:


marco-jantke:
Can't tell about the implementation but we'd be looking forward for this kind of functionality as well. Therefore, thumbs up :-)

PR #147, opened 2019-04-01
%!S seen when using console writer with NoLevel

dcarbone:
Currently when using the provided console logger with a level of `NoLevel`, you get this sort of output:
`"<nil> %!S Foobar bar=true foo=[1,2,3]\n"`.  I think ideally this should be `"<nil> Foobar bar=true foo=[1,2,3]\n"`

I have created a test case and implemented a proposed fix for this.

### DIFF
--- a/console.go
+++ b/console.go
@@ -321,6 +321,8 @@ func consoleDefaultFormatLevel(noColor bool) Formatter {
 			default:
 				l = colorize("???", colorBold, noColor)
 			}
+		} else if i == nil {
+			l = ""
 		} else {
 			l = strings.ToUpper(fmt.Sprintf("%s", i))[0:3]
 		}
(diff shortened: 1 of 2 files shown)


console.go
```diff
@@ -321,6 +321,8 @@ func consoleDefaultFormatLevel(noColor bool) Formatter {
 			default:
 				l = colorize("???", colorBold, noColor)
 			}
+		} else if i == nil {
+			l = ""
```
dcarbone: I'm not certain of all the possible side-effects of this, but i would presume its safe enough if the input is nil, its probably preferable to not try to treat it as a string and just move on.

dcarbone:
This is no longer necessary as of 1.14.

PR #158, opened 2019-06-08
Add TraceLevel

crazy-max:
Sometimes you want finer-grained informational events and need to debug something that outputs way too much data to log outside of a specific build when you're targeting that particular thing and do not care about errors or other logging info (since the volume of trace info will obscure them).

I think this could encroach on loggers like journald and that's why I'm asking for your opinion on adding this new log level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ func main() {
 * warn (`zerolog.WarnLevel`, 2)
 * info (`zerolog.InfoLevel`, 1)
 * debug (`zerolog.DebugLevel`, 0)
-* trace (`zerolog.TraceLevel`, -1)
+* trace (`zerolog.TraceLevel`, 0)
 
 You can set the Global logging level to any of these options using the `SetGlobalLevel` function in the zerolog package, passing in one of the given constants above, e.g. `zerolog.InfoLevel` would be the "info" level.  Whichever level is chosen, all logs with a level greater than or equal to that level will be written. To turn off logging entirely, pass the `zerolog.Disabled` constant.
 
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ func main() {
 * warn (`zerolog.WarnLevel`, 2)
 * info (`zerolog.InfoLevel`, 1)
 * debug (`zerolog.DebugLevel`, 0)
+* trace (`zerolog.TraceLevel`, -1)
 
 You can set the Global logging level to any of these options using the `SetGlobalLevel` function in the zerolog package, passing in one of the given constants above, e.g. `zerolog.InfoLevel` would be the "info" level.  Whichever level is chosen, all logs with a level greater than or equal to that level will be written. To turn off logging entirely, pass the `zerolog.Disabled` constant.
 
--- a/README.md
+++ b/README.md
@@ -114,13 +114,13 @@ func main() {
 
 **zerolog** allows for logging at the following levels (from highest to lowest):
 
-* panic (`zerolog.PanicLevel`, 6)
-* fatal (`zerolog.FatalLevel`, 5)
-* error (`zerolog.ErrorLevel`, 4)
-* warn (`zerolog.WarnLevel`, 3)
-* info (`zerolog.InfoLevel`, 2)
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -120,6 +120,7 @@ func main() {
 * warn (`zerolog.WarnLevel`, 2)
 * info (`zerolog.InfoLevel`, 1)
 * debug (`zerolog.DebugLevel`, 0)
+* trace (`zerolog.TraceLevel`, 0)
```
crazy-max: Don't really know how to handle an untyped int for this new level.

lammel: Probably it makes sense to shift all log level + 1 and start with trace.
Question is if the actual number is used directly by applications using zerolog.
Usually you should use the constants when referring to specific levels.

So Trace = 0, Debug = 1, ... should not impact to many, but can be considered a breaking change.

lammel:
We definitely need a trace level for logging in some libraries where very detailed information is needed for troubleshooting.
E.g. Logging of HTTP Response would be DEBUG level, but the acutal body (which could be large and clutter logs) would be logged on TRACE level.

Currently we work around the missing trace level, by using an additional boolean in each lib, which tells it to log additional information, which is rather clunky to use.

Even when having logging per section (see #67) it still makes sense to have a trace level for fine grained information.

README.md
crazy-max: Exactly. I have started this feature by shifting all log level like you said but I thought it would have a big impact on applications that use it directly.
And yeah we could mark this as breaking change :)

lammel:
Is there any chance that the additional TraceLevel is going to be merged (or the filtering in issue #67)?
A short heads up on that would be greatly appreciated.

vladyslav2:
I have same question, any ideas will that pull request will be excepted or not?
@crazy-max  thanks for your work

rs:
I'm hesitant on that one because it is a behavior breaking change if one relies on current numeric values for levels. We could go to the v2 path, but I would love to avoid that.

vladyslav2:
@rs Sorry, I might be ignorant, but if you using numerics - you are kind of expecting things to break, no?
Will you accept a pull request with `-1` for trace?

rs:
Ok for -1.

crazy-max:
@vladyslav2 I can do that on my branch if you want?

rs:
@crazy-max you missed `log_test.go`

crazy-max:
@rs Yes and also `Level` is an unsigned integer and it breaks "a lot" of things

crazy-max:
@rs Not sure if my last commit is the good strategy here :confused:

crazy-max:
@rs Should be ok now :slightly_smiling_face:

log_test.go
```diff
@@ -508,30 +508,34 @@ func TestLevelWriter(t *testing.T) {
...
 		{DebugLevel, `{"level":"debug","message":"1"}` + "\n"},
 		{InfoLevel, `{"level":"info","message":"2"}` + "\n"},
 		{WarnLevel, `{"level":"warn","message":"3"}` + "\n"},
 		{ErrorLevel, `{"level":"error","message":"4"}` + "\n"},
 		{NoLevel, `{"message":"nolevel-1"}` + "\n"},
-		{DebugLevel, `{"level":"debug","message":"5"}` + "\n"},
-		{InfoLevel, `{"level":"info","message":"6"}` + "\n"},
-		{WarnLevel, `{"level":"warn","message":"7"}` + "\n"},
-		{ErrorLevel, `{"level":"error","message":"8"}` + "\n"},
+		{TraceLevel, `{"level":"trace","message":"5"}` + "\n"},
```
rs: Duplicated

rs:
When my last comment will be resolved.

log_test.go
rs: You're right, I misread.

crazy-max: Ok thanks!

mleonix:
Thank you all!

vladyslav2:
@crazy-max thanks man!

PR #164, opened 2019-07-02
add send function for convenience

i6du:


event.go
```diff
@@ -105,6 +105,11 @@ func (e *Event) Msg(msg string) {
 	e.msg(msg)
 }
 
+// End call Msg with empty string.
+func (e *Event) End() {
```
rs: What about `Send` instead?

rs:
Could you also update the README to mention it?

i6du:
updated

PR #165, opened 2019-07-05
Simplify code of Event.msg(msg string)

SenZhangAI:
Simplify the code of `Event.msg(msg string)`,

The test passed after modification.

I also test if it  has an impact on the speed by running `BenchmarkHooks` for three times:

Test Environment:
```
os: darwin
arch: amd64
version: go 1.12.6
```

Before modified:

```
BenchmarkHooks/Nop/Single-4         	50000000	        40.4 ns/op
BenchmarkHooks/Nop/Multi-4          	30000000	        50.8 ns/op
BenchmarkHooks/Simple-4             	20000000	        77.2 ns/op

BenchmarkHooks/Nop/Single-4         	30000000	        40.4 ns/op
BenchmarkHooks/Nop/Multi-4          	30000000	        46.2 ns/op
BenchmarkHooks/Simple-4             	20000000	        78.0 ns/op

BenchmarkHooks/Nop/Single-4         	50000000	        42.1 ns/op
BenchmarkHooks/Nop/Multi-4          	30000000	        45.8 ns/op
BenchmarkHooks/Simple-4             	20000000	        74.9 ns/op
```

After modified:

```
BenchmarkHooks/Nop/Single-4         	30000000	        38.5 ns/op
BenchmarkHooks/Nop/Multi-4          	30000000	        47.2 ns/op
BenchmarkHooks/Simple-4             	20000000	        75.0 ns/op

BenchmarkHooks/Nop/Single-4         	30000000	        39.9 ns/op
BenchmarkHooks/Nop/Multi-4          	30000000	        45.3 ns/op
BenchmarkHooks/Simple-4             	20000000	        77.2 ns/op

BenchmarkHooks/Nop/Single-4         	50000000	        44.3 ns/op
BenchmarkHooks/Nop/Multi-4          	30000000	        44.0 ns/op
BenchmarkHooks/Simple-4             	20000000	        76.4 ns/op
```

### DIFF
--- a/event.go
+++ b/event.go
@@ -127,13 +127,8 @@ func (e *Event) Msgf(format string, v ...interface{}) {
 }
 
 func (e *Event) msg(msg string) {
-	if len(e.ch) > 0 {
-		e.ch[0].Run(e, e.level, msg)
-		if len(e.ch) > 1 {
-			for _, hook := range e.ch[1:] {
-				hook.Run(e, e.level, msg)
-			}
-		}
+	for _, hook := range e.ch {
+		hook.Run(e, e.level, msg)
 	}
 	if msg != "" {
 		e.buf = enc.AppendString(enc.AppendKey(e.buf, MessageFieldName), msg)


rs:
Doesn't this optimize events with hooks to the expense of events with no hooks or a single hook?

SenZhangAI:
Thank you ! 
I got your idea now, I was confused by this code and didn't know the reason of writing like this.

My simple test may show it may has no significant impact on optimizing events with no hooks or a single hook.  

But I don't know if it will work in other go versions or other environments, I will check the result in travis when I am not busy :)

smyrman:
You can inspect the generated Asembly by issuing:
```
go build -o test .
go tool objdump test > test.s

**Comment by smyrman** [comment]
I am not an expert in neither Assembly nor assembly operation performance. Why do you belive the original code to be faster with no or one hook @rs?

I am asking mainly out of coriousity as I to was confused by the original code 😄

**Comment by SenZhangAI** [comment]
@smyrman 
Thank you for your advise. I will try it in weekend :)

**Comment by smyrman** [comment]
**Disclaimer: The benchmark results and code readability is more important than the content of this comment. This is mostly a procrastination from my side.**

Again, this is _not_ my expertise, but here is my very brief assessment on the performance based on the Assembly.

# Brief analysis

The `if` causes an _conditional jump_ (`CMPQ`, `JLE`), as seen in the original code:
```asm
  event.go:130		0x40f14			4883784800		CMPQ $0x0, 0x48(AX)		
  event.go:130		0x40f19			0f8ed3000000		JLE 0x40ff2
```
The for-loop causes one _unconditional jump_ (`JMP`), and one conditional jump (`CMPQ`, `JL`):
```asm
  event.go:133		0x40f84			eb17			JMP 0x40f9d
  ...
  event.go:133		0x40fed			4839c8			CMPQ CX, AX			
  event.go:133		0x40ff0			7c94			JL 0x40f86
```

This gives the following number of conditional jumps in the execution path based on number of hooks:

- 0 hooks:
  - original code: 1 conditional jump
  - new code: 1 conditional jump
- 1 hook:
  - original code: 2 conditional jump
  - new code: 2 conditional jump
- 2 hook:
  - original code: 3 conditional jumps
  - new code: 3 conditional jumps

# Reasoning

It's my general understanding that an unconditional `JMP`, is [more or less free](https://stackoverflow.com/questions/5127833/meaningful-cost-of-the-jump-instruction) on a modern CPU.

A conditional jump is a _branch condition_, which means it affects the CPU pipeline, and could have much higher potential cost. The pipeline will be pre-filled with instructions for the CPU to execute in order or in parallel. Modern CPUs have good branch prediction to cope with this and can sometimes compute multiple "branches" at once for later to discard the branch that wasn't chosen. However the more conditional branches you get, the higher is the chance of the wrong branch being predicted, and that has some cost. Another expensive operation, is to access the main memory, which we probably need to do if we want to actually call a hook. Doing so unnecessary could waste a lot of cycles.

# Conclusion

My conclusion from that is that it's hard for me to reason _why_ or _how_ the original code (at least on Mac and compiled with Go 1.13) can be faster than your code @SenZhangAI, as they have the same number of conditional branches. This basically supports your findings from the benchmarks, that there is _no significant performance difference_.

In that case, _I_ would prefer the cleaner code.

On the other hand, I might be missing something, or the generated assembly may be different on another platform. If there _is clear evidence_ that the existing code is more performant, then a comment explaining _why_ the code is written as it is should be considered.

rs:
Check the all bench, you're right, it makes no difference. I guess this trick use to do something with an older version of Go.

PR #177, opened 2019-08-13
Let user decide what to do if logger is not in context

tomaszczerminski:
As for now disabledLogger is being returned if user tries to get logger from context.Context.
Such a behaviour may be beneficial, but in other cases it will be harmful.
Giving user a control over disabledLogger behavior would be beneficial.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -4,11 +4,11 @@ import (
 	"context"
 )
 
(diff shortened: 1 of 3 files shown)


mitar:
Isn't there already `DefaultContextLogger`?

PR #179, opened 2019-08-20
make sure event.Stack() is called before event.Err()

larsla:
This fixes the issue where ```log.Logger = log.With().Stack().Logger()``` does not work due to the hooks being called after ```log.Error().Err(err)```.


context.go
```diff
@@ -399,6 +399,7 @@ var sh = stackTraceHook{}
 // Stack enables stack trace printing for the error passed to Err().
 func (c Context) Stack() Context {
 	c.l = c.l.Hook(sh)
```
sergeydobrodey, line 401: This line and `type stackTraceHook struct{}` should be deleted

PR #186, opened 2019-10-03
Add Stringer and Stringers method for *Event  context

dennypenta:
Hi.
I made an implementation for adding values as a fmt.Stringer interface. It allows don't create new strings from struct in many cases and do it only if your log level and other conditions require it.

### DIFF
--- a/event.go
+++ b/event.go
@@ -241,6 +241,10 @@ func (e *Event) Stringer(key string, val fmt.Stringer) *Event {
 	if e == nil {
 		return e
 	}
+	if val == nil {
+		e.buf = enc.AppendString(enc.AppendKey(e.buf, key), "nil")
+		return e
+	}
 	e.buf = enc.AppendString(enc.AppendKey(e.buf, key), val.String())
 	return e
 }
--- a/event.go
+++ b/event.go
@@ -235,6 +235,26 @@ func (e *Event) Strs(key string, vals []string) *Event {
 	return e
 }
 
+// Stringer adds the field key with val as a string to the *Event context.
+// Use Stringer if possible, because it creates string value in lazy mode.
+func (e *Event) Stringer(key string, val fmt.Stringer) *Event {
+	if e == nil {
+		return e
+	}
+	e.buf = enc.AppendString(enc.AppendKey(e.buf, key), val.String())
+	return e
+}
+
+// Stringers adds the field key with vals as a []string to the *Event context.
+// Use Stringer if possible, because it creates string values in lazy mode.
+func (e *Event) Stringers(key string, vals []fmt.Stringer) *Event {
+	if e == nil {
+		return e
+	}
+	e.buf = enc.AppendStringers(enc.AppendKey(e.buf, key), vals)
+	return e
+}
+
 // Bytes adds the field key with val as a string to the *Event context.
 //
 // Runes outside of normal ASCII ranges will be hex-encoded in the resulting
--- a/event.go
+++ b/event.go
@@ -242,7 +242,7 @@ func (e *Event) Stringer(key string, val fmt.Stringer) *Event {
 		return e
 	}
 	if val == nil {
-		e.buf = enc.AppendString(enc.AppendKey(e.buf, key), "nil")
+		e.buf = enc.AppendInterface(enc.AppendKey(e.buf, key), nil)
 		return e
 	}
 	e.buf = enc.AppendString(enc.AppendKey(e.buf, key), val.String())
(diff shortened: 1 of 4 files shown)


rs:
I think the `Interface` would have the same effect.

event.go
```diff
@@ -235,6 +235,26 @@ func (e *Event) Strs(key string, vals []string) *Event {
 	return e
 }
 
+// Stringer adds the field key with val as a string to the *Event context.
+// Use Stringer if possible, because it creates string value in lazy mode.
+func (e *Event) Stringer(key string, val fmt.Stringer) *Event {
+	if e == nil {
+		return e
+	}
+	e.buf = enc.AppendString(enc.AppendKey(e.buf, key), val.String())
```
hummerd: This will cause panic if val is nil

hummerd:
> I think the `Interface` method would have the same effect.
```
log.Interface("url", url) // will produce {"url":{"Scheme":"http","Opaque":"","User":null,"Host":"example.com","Path":"/some/path","RawPath":"","ForceQuery":false,"RawQuery":"","Fragment":""}}

// While 
log.Stringer("url", url) // will produce {"url": "http://example.com/some/path"}
```

event.go
dennypenta: thanks, my miss

dennypenta: pushed the fix

dennypenta:
If you ok with this feature, I'm going to write tests

dennypenta:
I added tests, but it seems like tests for tests, probably it will be better to remove them.

event.go
```diff
@@ -235,6 +235,30 @@ func (e *Event) Strs(key string, vals []string) *Event {
...
 }
 
+// Stringer adds the field key with val as a string to the *Event context.
+// Use Stringer if possible, because it creates string value in lazy mode.
+func (e *Event) Stringer(key string, val fmt.Stringer) *Event {
+	if e == nil {
+		return e
+	}
+	if val == nil {
+		e.buf = enc.AppendString(enc.AppendKey(e.buf, key), "nil")
```
rs: Should be “null”.

internal/json/string_test.go
```diff
@@ -72,6 +100,15 @@ func TestAppendString(t *testing.T) {
...
 }
 
+func TestAppendStringer(t *testing.T) {
+	for _, tt := range encodeStringerTests {
+		b := enc.AppendString([]byte{}, tt.in.String())
+		if got, want := string(b), tt.out; got != want {
+			t.Errorf("appendString(%q) = %#q, want %#q", tt.in.String(), got, want)
+		}
+	}
+}
```
rs: Please add a test for nil case.

event.go
hummerd: You can use e.buf = enc.AppendInterface(enc.AppendKey(e.buf, key), nil) to achieve that

internal/json/string.go
```diff
@@ -29,6 +32,33 @@ func (e Encoder) AppendStrings(dst []byte, vals []string) []byte {
...
+	if vals[0] == nil {
+		dst = e.AppendInterface(dst, nil)
+	} else {
+		dst = e.AppendString(dst, vals[0].String())
+	}
+
+	if len(vals) > 1 {
+		for _, val := range vals[1:] {
+			if val == nil {
+				dst = e.AppendString(append(dst, ','), "nil")
```
rs: `nil` is not valid in json.

dennypenta:
Guys, I made an update month ago.
If I need a little bit fixes - let me know pls.

event.go
```diff
@@ -235,6 +235,30 @@ func (e *Event) Strs(key string, vals []string) *Event {
...
+// Use Stringer if possible, because it creates string value in lazy mode.
+func (e *Event) Stringer(key string, val fmt.Stringer) *Event {
+	if e == nil {
+		return e
+	}
+	if val == nil {
+		e.buf = enc.AppendInterface(enc.AppendKey(e.buf, key), nil)
+		return e
+	}
+	e.buf = enc.AppendString(enc.AppendKey(e.buf, key), val.String())
```
rs: Either this implementation should go to an `AppendStringer` or `AppendStringers` should be moved in this file. It is odd to have this feature implemented in two places.

dennypenta: I'm not sure what are you talking about. 
You propose to implement `AppendStriner` for encoder, but it will whole copy of `AppendString` . Or I can call `AppendString ` inside the `AppendStriner `

dennypenta:
Good day.
Guys, what is the plan?

event.go
rs: My point is that you repeat the `if val == nil then AppendInterface else AppendString` in many places where it should be a `AppendStringer` in the encoders instead.

PR #189, opened 2019-10-11
Make TimeFormatUnixMicro global var exportable

hazimj:
This PR patches a typo in https://github.com/rs/zerolog/pull/188 to make `zerolog.TimeFormatUnixMicro` exportable.


wphan:
does `timeFormatUnixMicro` need to be taken into account in `AppendTimes` as well?

https://github.com/hazimj/zerolog/blob/fix-TimeFormatUnixMicro/internal/json/time.go#L31

rs:
Good catch. @hazimj can you please add this?

PR #190, opened 2019-10-13
[journald] Writter should return written size when message was sent without error

IljaN:
The journald writer currently always returns zero. This causes an ErrShortWrite if used together with MultiWriter: https://github.com/golang/go/blob/master/src/io/multi.go#L64
- Fixes ErrShortWrite if used with io.MultiWriter

### DIFF
--- a/journald/journald.go
+++ b/journald/journald.go
@@ -71,7 +71,9 @@ func (w journalWriter) Write(p []byte) (n int, err error) {
 		err = fmt.Errorf("cannot connect to journalD")
 		return
 	}
+
 	var event map[string]interface{}
+	origPLen := len(p)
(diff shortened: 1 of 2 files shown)


rs:
Tests are failing. Can you please fix them so I can merge?

IljaN:
Not sure, need to debug, they pass locally.

rs:
Try with `go test -tags binary_log`

plar:
@rs Could you merge that PR into the master please? Tests have been fixed.

```
github.com/IljaN/zerolog(master)$ go test -tags binary_log
PASS
ok  	github.com/rs/zerolog	0.011s
```

PR #192, opened 2019-10-18
Event.Merge

ChieveiT:
I think it will be more usual to merge a dict rather than add a dict as a field of event. Flat structure is preferred in logging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,15 @@ log.Info().
     ).Msg("hello world")
 
 // Output: {"level":"info","time":1494567715,"foo":"bar","dict":{"bar":"baz","n":1},"message":"hello world"}
+
+log.Info().
+    Str("foo", "bar").
+    Merge(zerolog.Dict().
+        Str("bar", "baz").
+        Int("n", 1),
+    ).Msg("hello world")
+
+// Output: {"level":"info","time":1494567715,"foo":"bar","bar":"baz","n":1,"message":"hello world"}
 ```
 
 ### Customize automatic field names
--- a/event.go
+++ b/event.go
@@ -166,6 +166,21 @@ func (e *Event) Dict(key string, dict *Event) *Event {
 	return e
 }
 
+// Merge a dict to the event context.
+// Use zerolog.Dict() to create the dictionary.
+func (e *Event) Merge(dict *Event) *Event {
+	if e == nil {
+		return e
+	}
+	if len(dict.buf) > 1 && dict.buf[0] == '{' {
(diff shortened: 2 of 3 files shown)


rs:
Can you explain the use case? Taking your readme example:

```
log.Info().
    Str("foo", "bar").
    Merge(zerolog.Dict().
        Str("bar", "baz").
        Int("n", 1),
    ).Msg("hello world")
```

Why would one do that instead of:

```
log.Info().
    Str("foo", "bar").
    Str("bar", "baz").
    Int("n", 1).
    Msg("hello world")
```

ChieveiT:
@rs There is one case that we `should` pass some log fields as a struct through functions and finally log them in the proper place, especially in error handling stack, some codes like:
```
type customError struct {
    Fields *zerolog.Event
    Err error
}

func emitError() error {
    // some codes return an err
    return &customError{
        log.Error().Str('foo', 'bar').Int('bar', 123),
        err,
    }
}

func handleAndLogError() {
    err := emitError()
    if err != nil {
        if e, ok := err.(*customError); ok {
            log.Panic().
                Merge(e.Fields).
                Str('we should log here', 'to log error once and avoid multiple annoying error logs').
                Send()
        }
    }
}
``` 
I'm not sure whether it is a best practice in golang error handling. It looks so hacky. Many people might log twice in emitError() and handleAndLogError(), which looks worse for me. So the code above  is the only one way I have found I can get the clear error logs.

rs:
It does not look right. You created two event but only emit one. Why don't you emit the original event, adding fields on it directly before sending?

ChieveiT:
I don't think I can directly send the original event. One error occurring should not emit two error events, that's why I need Event.Merge to collect the error context through call stack and finally send them once.

rs:
I would do it this way:

```go
type customError struct {
    Foo string
    Bar int
    Err error
}

func (e customError) Error() string {
	return e.Err.Error()
}

func (e customError) AddLogFields(e *zerolog.Event) *zerolog.Event {
    return e.Err(e).
        Str('foo', e.Foo).
        Int('bar', e.Bar)
}

func emitError() error {
    // some codes return an err
    return &customError{
        Foo: "bar",
        Bar: 123,
        Err: err,
    }
}

func handleAndLogError() {
    err := emitError()
    if err != nil {
        if e, ok := err.(*customError); ok {
            e.AddLogFields(log.Panic()).
                Err(e.Err).
                Str('we should log here', 'to log error once and avoid multiple annoying error logs').
                Send()
        }
    }
}
```

PR #196, opened 2019-10-30
modify Caller

guonaihong:
### Why write this pr
 To modify the Caller, you can pass a jump to a few function call stacks.
### demo
```go
    package main
    
    import (
        "github.com/rs/zerolog"
        "github.com/rs/zerolog/log"
    )
    
    func myError() {
        log.Debug().Caller(1).Str("test2", "v2").Send()
    }
    
    func foo() {
        log.Debug().Caller(2).Str("test2", "v2").Send()
    }
    
    func myError2() {
        foo()
    }
    func main() {
        zerolog.TimeFieldFormat = zerolog.TimeFormatUnix
    
        log.Debug().Caller().Str("test1", "v1").Send()
        myError()
        myError2()
    }
```

### DIFF
--- a/event.go
+++ b/event.go
@@ -665,8 +665,14 @@ func (e *Event) Interface(key string, i interface{}) *Event {
 }
 
 // Caller adds the file:line of the caller with the zerolog.CallerFieldName key.
-func (e *Event) Caller() *Event {
-	return e.caller(CallerSkipFrameCount)
+func (e *Event) Caller(skip ...int) *Event {
+
+	sk := CallerSkipFrameCount
+	if len(skip) > 0 {
+		sk = skip[0] + CallerSkipFrameCount
+	}
+
+	return e.caller(sk)
 }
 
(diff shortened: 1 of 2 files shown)


event.go
```diff
@@ -665,8 +665,14 @@ func (e *Event) Interface(key string, i interface{}) *Event {
 }
 
 // Caller adds the file:line of the caller with the zerolog.CallerFieldName key.
```
rs: Can you document the new argument please?

go.mod
```diff
@@ -7,3 +7,4 @@ require (
 	github.com/zenazn/goji v0.9.0
 	golang.org/x/tools v0.0.0-20190828213141-aed303cbaa74
 )
+
```
rs: Did go mod add this?

guonaihong: Will be deleted here

event.go
guonaihong: Modify now

PR #201, opened 2019-11-29
add context.FieldsWith

exherb:


rs:
I don’t see a use-case for that. Can you give one?

exherb:
some log system like graylog use _ as special mark https://docs.graylog.org/en/3.1/pages/gelf.html

rs:
I'm not sure it is the role of the log library to add those. It' easy enough to do that in your app.

exherb:
then chain is not possible …

rs:
How about that?

```go
func fieldsPrefix(fields map[string]interface{}, prefix string) map[string]interface{} {
 	r := make(map[string]interface{}, len(fields))
 	for k, v := range fields {
 		r[strings.Join([]string{prefix, k}, "")] = v
 	}
 	return r
 }

log.Info().Fields(fieldsPerfix(someFields, "_")).Msg("Hello")
```

exherb:
ok, thanks.

PR #216, opened 2020-03-07
Add context hook support

ndk:
I've got an idea to retranslate logger messages to Opentracing. It needs to add some mechanics to be able to propagate the span context from context.Context to the logger.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -42,6 +42,10 @@ func (l *Logger) WithContext(ctx context.Context) context.Context {
 // is associated, a disabled logger is returned.
 func Ctx(ctx context.Context) *Logger {
 	if l, ok := ctx.Value(ctxKey{}).(*Logger); ok {
+		for _, ctxHook := range l.ctxHooks {
+			l2 := ctxHook.Run(ctx, *l)
+			l = &l2
+		}
 		return l
 	}
 	return disabledLogger
--- a/ctx_hook.go
+++ b/ctx_hook.go
@@ -1,0 +1,17 @@
+package zerolog
+
+import "context"
+
+// Hook defines an interface to a contex hook.
+type CtxtHook interface {
+	// Run runs the hook with extracting of the Logger associated with the ctx.
+	Run(ctx context.Context, l Logger) Logger
+}
+
+// CtxtHookFunc is an adaptor to allow the use of an ordinary function as a Hook.
+type CtxtHookFunc func(ctx context.Context, l Logger) Logger
+
+// Run implements the CtxtHook interface.
+func (h CtxtHookFunc) Run(ctx context.Context, l Logger) Logger {
+	return h(ctx, l)
+}
--- a/log.go
+++ b/log.go
@@ -183,11 +183,12 @@ func ParseLevel(levelStr string) (Level, error) {
 // serialization to the Writer. If your Writer is not thread safe,
 // you may consider a sync wrapper.
 type Logger struct {
-	w       LevelWriter
-	level   Level
-	sampler Sampler
-	context []byte
-	hooks   []Hook
+	w        LevelWriter
+	level    Level
+	sampler  Sampler
+	context  []byte
+	hooks    []Hook
+	ctxHooks []CtxtHook
 }
 
 // New creates a root logger with given output writer. If the output writer implements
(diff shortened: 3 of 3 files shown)


rs:
I need more context to understand how this work and how / for what it would be used. Note that I'm not familiar with Opentracing.

ndk:
Yeah, sure. First of all, this is inspired by this link. Unfortunately, it wouldn't work with Jaeger.

(https://www.elastic.co/guide/en/apm/agent/go/current/builtin-modules.html#builtin-modules-apmzerolog).

Like the logger, a span context is stored in 'context.Context'. So, it's just enough to extract a span context from 'context.Context' at the same moment when we're extracting the logger. Then we could put it into the logger like yet another field. In this manner, we could propagate a span context to the logger writer.

It gives us an ability to retranslate the logger events to Opentracing.

I've made PoC. If you wish I can demonstrate to you how it works

byamba-tumurkhuu:
I have a use case that can benefit from this. We're trying to attach correlation IDs to our logs.  The IDs would be context so we'd need hooks that work with context.  Here is the rough idea of how it'd work:

```go
// CorrelationIdHook is a CtxtHook that extracts correlation IDs from ctx and attaches to the log
type CorrelationIdHook struct{}


func (h CorrelationIdHook) Run(ctx context.Context, e *zerolog.Event) Logger {
        return e.Str("correlationId", extractId(ctx))
}

// Register context hook
log.logger = log.CtxHook(CorrelationIdHook{})

// Usage
log.Ctx(ctx).Info().Msg("File processed")
```

Notice the `Run` method has a slightly different interface than what's suggested. It's would be just a regular hook that allows users to interact with the context.

rs:
Why don't you just do:

```
log.Info().Str("correlationId", extractId(ctx)).Msg("File processed")
```

byamba-tumurkhuu:
Yes, we can do that. We have more context values that need to be logged than just the correlation IDs. We can also have another function that takes the logger and decorates with these values. But it'd be much more convenient to solve this with a hook.

byamba-tumurkhuu:
If you have to pass additional configuration parameters to that function it becomes inconvenient:
```go
logger := GetTracingLogger(ctx, config1, config2)
logger.Info().Msg("...")
```

The code that was just trying to write some log now needs to be aware of these configurations. It feels like a good use of a  hook except the hooks don't support context.

rs:
It would be:

```
TraceLog(logger.Info(), ctx, config1, config2).Msg("")
```

byamba-tumurkhuu:
Yes, the issue here is I don't want some random code of mine to know about `config1` and `config2`. They should be doing just `TraceLog(logger.Info(), ctx).Msg("")`. I'd like to configure `config1` and `config2` when I'm configuring other global logging variables and forget they exist.

rs:
So put then in the context and extract them in `TraceLog`.

byamba-tumurkhuu:
Then I'll have to set these config values (they're static) for every user request context.

byamba-tumurkhuu:
A little clarification, they're static in a sense that they get their value when application starts like from environment variable and don't change.

rs:
So create a lambda or a type with those values.

rs:
If they never change, they can also be part of the context.

byamba-tumurkhuu:
I just want to say I appreciate your suggestions and responsiveness. With that said creating a struct or something similar that holds these configurations definitely helps. Now the caller needs to know about just that struct regardless of how many configurations it holds. 

However, the issue is still there, the caller still needs to know about that specific instance of the struct (through function param, global variable etc) which is pretty much irrelevant to what the callers do.

rs:
You can pass that struct as your logger instead of a zerolog logger. That struct can wrap the logger and do whatever you want. The point is, I don't want to add complexity to zerolog in order to address all custom needs, custom needs should be address in your code.

uynap:
> I need more context to understand how this work and how / for what it would be used. Note that I'm not familiar with Opentracing.



> You can pass that struct as your logger instead of a zerolog logger. That struct can wrap the logger and do whatever you want. The point is, I don't want to add complexity to zerolog in order to address all custom needs, custom needs should be address in your code.

I'm sorry to dig up history comments. I agree with your point. However, Opentelementry/Opentracing is NOT a custom need. It's a widely used technology that's needed to be taken into consideration by a logging library.

Log tracing is a MUST feature in microservice world and Opentracing is the mainstream solution in this area. This context hook can save all people's needs who're using zerolog with Opentracing. https://github.com/opentracing/opentracing-go

rs:
Sorry but I fail to see what is missing to integrate zerolog with opentracing. All the hooks should be in place already.

PR #217, opened 2020-03-11
export buf

evgenigourvitch:
While using hook capability, in my use case, I had to access buf member in zerolog.Event struct, which is currently available only using reflection, which affects the performance.

### DIFF
--- a/event.go
+++ b/event.go
@@ -79,6 +79,15 @@ func (e *Event) write() (err error) {
 	return
 }
 
+// Export buf member which casn be useful in some cases
+// while using hooks capability instead of using reflection
+func (e *Event) GetBuffer() []byte {
+	if e == nil {
+		return nil
+	}
+	return e.buf
+}
+
 // Enabled return false if the *Event is going to be filtered out by
 // log level or sampling.
 func (e *Event) Enabled() bool {


rs:
This is quite dangerous and won't be compatible with the binary serialization mode. Can you give me more detail on your use-case?

eugenepx:
> This is quite dangerous and won't be compatible with the binary serialization mode. Can you give me more detail on your use-case?

Thanks for the response,

Regarding the use case - it is pretty simple - in Run method (required by https://github.com/rs/zerolog/blob/master/hook.go#L6) the e variable contains the data which was previously appended in the main app like log.Info().Str("version", version).Str("addr", conf.Addr) and so on (in some cases it contains serialised objects) and at this point I need an access to it to do several things with this raw data (as slice of bytes) and getting buf variable itself could be extremely useful for me. At this point I have to use a reflection, like:
```golang
fv := reflect.ValueOf(e).Elem().FieldByName("buf")
if !fv.IsValid() {
//    handle error and return
}
do some stuff with fv.Bytes()
```
Which has some performance penalties, which could be avoided if a buf variable will be exportable.
I think other people in the community also could face with a similar use case, that's why this PR.
Thanks.

evgenigourvitch:
> This is quite dangerous and won't be compatible with the binary serialization mode. Can you give me more detail on your use-case?

Thanks for the response,

Regarding the use case - it is pretty simple - in Run method (required by https://github.com/rs/zerolog/blob/master/hook.go#L6) the e variable contains the data which was previously appended in the main app like log.Info().Str("version", version).Str("addr", conf.Addr) and so on (in some cases it contains serialised objects) and at this point I need an access to it to do several things with this raw data (as slice of bytes) and getting buf variable itself could be extremely useful for me. At this point I have to use a reflection, like:
```golang
fv := reflect.ValueOf(e).Elem().FieldByName("buf")
if !fv.IsValid() {
//    handle error and return
}
do some stuff with fv.Bytes()
```
Which has some performance penalties, which could be avoided if a buf variable will be exportable.
I think other people in the community also could face with a similar use case, that's why this PR.
Thanks.

rs:
I'm really not comfortable exposing an internal structure. This will prevent us from making any future evolution of how the log event is created.

evgenigourvitch:
It is exposed right now, de facto (via reflection) just wanted to make it more efficient and more... user friendly, since only read method is about to be added and no modification could be done to the buf itself via it.
But you're the master :)
Thanx anyway!

rs:
By using reflexion you know that what you are doing is not supported / dangerous. Exposing something in the APi means the behavior is supported and won’t break in the future. It is a major difference.

evgenigourvitch:
Yeah, I got you...
Can you name, please, any use case when exposing buf member will cause any issue?
Thinking about any, but don't think I can see any at this point...
Thanks!

rs:
* You change the content of the buffer, the code expects the buffer to be a certain way and writes invalid json
* You keep a pointer to the buffer after it is put back in the sync.Pool
* We decide to refactor the JSON writer and the buffer is no longer available to us (forcing a non backward compat version or us not doing the refactor)

In general it is a design small to explose internals.

evgenigourvitch:
I understand... as for me, the most reasonable bullet is the 2nd one, but the design limitations also sounds reasonable...
Can you advise, please, the correct way to get the log data from the Run method (https://github.com/rs/zerolog/blob/master/hook.go#L6) while the message variable is empty because all the information which was dumped to log was supplied via public methods of the event like https://github.com/rs/zerolog/blob/master/event.go#L576 or https://github.com/rs/zerolog/blob/master/event.go#L378 and actual data is stored within, non-exported, buf variable and, personally, I don't like using of reflection due to performance penalty.

Thanks!

rs:
If you need the data added to the logger in your app, my advise would be to make this data available by other means. A logger isn’t a good way to pass data around in your app.

evgenigourvitch:
Understood.
Thanks a lot for you time!

PR #223, opened 2020-04-13
Remove dependency on Goji

shabbyrobe:
Hello, I noticed Zerolog was bringing Goji into my dependency graphs unnecessarily; only the `mutil` subpackage was being used, so I've vendored it in here (with the exception of the <=1.8 code) after noticing there was some precedence for this with the Diode library.

It's kind of a separate discussion, but I also notice that there's a systemd package that also shows up, but that will be trickier to manage as that seems to me like a dependency that you would very much not want to vendor! If the `mutil` change I've proposed here is agreeable, would you also be up for discussing some of the ways to manage the systemd dependency as well?


rs:
For Goji I was thinking about doing a sub-module, but the last time I tried, it was a PITA to manage. I don't know if it improved since then.

shabbyrobe:
Wow that was quick, thank you!

By "sub-module", do you mean a nested "go module", or a git submodule?

rs:
go module.

Issue #163, opened 2019-06-25
Int overflow on 32 bits arches


eclipseo:
Golang 1.12.6 on i686 and armv7:
```
Testing    in: /builddir/build/BUILD/zerolog-1.14.3/_build/src
         PATH: /builddir/build/BUILD/zerolog-1.14.3/_build/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/sbin
       GOPATH: /builddir/build/BUILD/zerolog-1.14.3/_build:/usr/share/gocode
  GO111MODULE: off
      command: go test -buildmode pie -compiler gc -ldflags "-X github.com/rs/zerolog/version=1.14.3 -extldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '"
      testing: github.com/rs/zerolog
github.com/rs/zerolog
PASS
ok  	github.com/rs/zerolog	0.033s
github.com/rs/zerolog/hlog
PASS
ok  	github.com/rs/zerolog/hlog	0.024s
github.com/rs/zerolog/internal/cbor
FAIL	github.com/rs/zerolog/internal/cbor [build failed]
BUILDSTDERR: # github.com/rs/zerolog/internal/cbor [github.com/rs/zerolog/internal/cbor.test]
BUILDSTDERR: ./types_test.go:91:3: constant 4294967294 overflows int
BUILDSTDERR: ./types_test.go:94:3: constant 3022355601293312 overflows int
BUILDSTDERR: ./types_test.go:95:3: constant 1000000000000 overflows int
BUILDSTDERR: ./types_test.go:119:3: constant -4294967294 overflows int
BUILDSTDERR: ./types_test.go:122:3: constant -3022355601293313 overflows int
BUILDSTDERR: ./types_test.go:123:3: constant -1000000000001 overflows int
```

toravir:
@eclipseo how do i recreate this ? i don't have an armv7 or i686 machine.. can u provide steps to create a 32-bit env pls ?

jsdidierlaurent:
You can reproduce this issue with:
`GOARCH=386 go test ./...`

toravir:
Thanks @jsdidierlaurent  - am able to recreate it - will looking into it and get back in a few days..

PR #230, opened 2020-05-12
fix for #163 

toravir:
Moved the 8byte integer tests to a different file and added a "!386" build tag on it.


toravir:
sorry some old commits got in..

PR #244, opened 2020-06-24
 Fix UpdateContext panic with empty Context

### LINKED ISSUE #239: Regression: panic when calling UpdateContext on a context not yet initialized
I'm seeing a new panic with [v0.19.0](https://github.com/rs/zerolog/releases/tag/v1.19.0) that I didn't experience with [v0.18.0](https://github.com/rs/zerolog/releases/tag/v1.18.0):

```
panic: runtime error: index out of range [-1]

goroutine 1 [running]:
github.com/rs/zerolog/internal/json.Encoder.AppendKey(...)
        /Users/hairyhenderson/go/pkg/mod/github.com/rs/zerolog@v1.19.0/internal/json/base.go:7
github.com/rs/zerolog.Context.Str(0x1144480, 0xc000098590, 0xff, 0x0, 0x0, 0xc0000cc400, 0x0, 0x1f4, 0x0, 0x0, ...)
        /Users/hairyhenderson/go/pkg/mod/github.com/rs/zerolog@v1.19.0/context.go:74 +0x202
main.main.func3(0x1144480, 0xc000098590, 0xff, 0x0, 0x0, 0xc0000cc400, 0x0, 0x1f4, 0x0, 0x0, ...)
        /Users/hairyhenderson/zlbug/main.go:21 +0xb8
github.com/rs/zerolog.(*Logger).UpdateContext(0xc0000d9e48, 0x1126520)
        /Users/hairyhenderson/go/pkg/mod/github.com/rs/zerolog@v1.19.0/log.go:255 +0xef
main.main()
        /Users/hairyhenderson/zlbug/main.go:20 +0x1b0
exit status 2
```

This code can reproduce it:

```go
package main

import (
	"os"

	"github.com/rs/zerolog"
)

func main() {
	log := zerolog.New(zerolog.ConsoleWriter{
		Out:             os.Stdout,
		NoColor:         true,
		FormatTimestamp: func(i interface{}) string { return "" },
		FormatLevel:     func(i interface{}) string { return "" },
	})

	log.Info().Msg("Hello world")
	log.UpdateContext(func(c zerolog.Context) zerolog.Context {
		return c.Str("foo", "bar")
	})
	log.Info().Msg("kaboom")
}
```

If I add a line like `log = log.With().Logger()` before calling `log.UpdateContext`, it's OK.

This was introduced by https://github.com/rs/zerolog/pull/226, specifically [here](https://github.com/rs/zerolog/pull/226/files#diff-7aba00861f27ecbc08d11f9626b6526eR7). The `dst` byte array is no longer being length-checked before attempting to access index `len(dst)-1` (`-1` when the array is empty - i.e. when it hasn't been initailized yet).

Note that I only ran into this in a unit test, and is easily worked around by calling `log.With()`, but it's reasonable to assume this will affect other code as well.

This workaround only works because `enc.AppendBeginMarker` [is now called](https://github.com/rs/zerolog/pull/226/files#diff-308aaa719425f4333033c97cfb3b9868R240) by `With`, and it tolerates a zero-length array and produces an array of length `1`.

/cc @ffenix113 @rs

ffenix113:
This fixes UpdateContext function when it is called with empty Context.

Fixes #239

### DIFF
--- a/log.go
+++ b/log.go
@@ -252,6 +252,9 @@ func (l *Logger) UpdateContext(update func(c Context) Context) {
 	if cap(l.context) == 0 {
 		l.context = make([]byte, 0, 500)
 	}
+	if len(l.context) == 0 {
+		l.context = enc.AppendBeginMarker(l.context)
+	}
 	c := update(Context{*l})
 	l.context = c.l.context
 }
--- a/log_test.go
+++ b/log_test.go
@@ -772,3 +772,19 @@ func TestErrorHandler(t *testing.T) {
 		t.Errorf("ErrorHandler err = %#v, want %#v", got, want)
 	}
 }
+
+func TestUpdateEmptyContext(t *testing.T) {
+	var buf bytes.Buffer
+	log := New(&buf)
+
+	log.UpdateContext(func(c Context) Context {
+		return c.Str("foo", "bar")
+	})
+	log.Info().Msg("no panic")
+
+	want := `{"level":"info","foo":"bar","message":"no panic"}` + "\n"
+
+	if got := buf.String(); got != want {
+		t.Errorf("invalid log output:\ngot:  %q\nwant: %q", got, want)
+	}
+}


hairyhenderson:
Thanks @ffenix113!

Issue #252, opened 2020-07-22
change log level dynamically


itapai:
Basically, title.

I want to be able to dynamically set the log level at runtime. Is this achievable with zerolog?
This can be done with zap. https://dev.to/procyclinsur/golang-dynamic-logging-3p7k

panlatent:
I need this feature because zero-log is used as the underlying library.

mpalmer:
I've achieved this in a HTTP service by setting the log level in the logger that is created for each request (which also carries all the per-request metadata like request ID).  This wouldn't work if you need to change log level mid-request, but I'm yet to find that a problem for a typical HTTP service, which deals with short-lived individual requests.

aldas:
For us would be use case - our application has API and background processes or shared services that log. During application start logger is created as very first thing. during application startup application configuration is also read in and it could/should be able to change log level.

rs:
Store the logger as a pointer with proper locking and change the logger as you please.

SEJeff:
In `go.uber.org/zap` this is handled with [AtomicLevel](https://pkg.go.dev/go.uber.org/zap?utm_source=godoc#AtomicLevel). The design of this is sensible and should be considered for zerolog.

achughZscaler:
Is there a plan to do this?

mitar:
I have made #583 to suggest that zerolog needs a batching post-processing capability for (possibly a batch of) log events before they are written out. This would allow one to decide the log level after observing what other log events have been generated for example during request handling. You could then log at debug log level but write the log line only if any of the log lines for the request was an error, or something like that.

avishaybp81:
What is wrong with calling `SetGlobalLevel` 
See https://goplay.tools/snippet/Qzy6O7eJEIW

PR #255, opened 2020-08-14
Add event level setter

panlatent:
Support #252 .

Zero-log's API first sets a level, which may not be the case in actual use. If you let go of this function, zero-log can support more, especially secondary packaging.


rs:
This will break level management. We start with level to decide if we need to create an event or not based on your current log level.

PR #278, opened 2021-01-03
Rewrite linter to use goanalysis package

tdakkota:
Package [`golang.org/x/tools/go/loader`](https://pkg.go.dev/golang.org/x/tools/go/loader) is deprecated. 
This PR uses special package [`golang.org/x/tools/go/analysis`](https://pkg.go.dev/golang.org/x/tools/go/analysis) for creating linters. It supports Go modules.

I tried to write a test using [`analysistest`](https://pkg.go.dev/golang.org/x/tools/go/analysis/analysistest), but Go modules in testdata are not supported for now (see https://github.com/golang/go/issues/37054).


rs:
Thanks for you PR. It does not seem to work on Go < 1.12. Can you please either fix it or exclude it from CI for Go < 1.12.

PR #282, opened 2021-01-18
Make MultiLevelWriter resilient to individual log failure

### LINKED ISSUE #281: MultiLevelWriter should not stop writing a message when one writer fails
MultiLevelWriter returns immediately as soon as any log writer returns an error. If you configure 2 writers and the first one fails  you get no output at all. This is very confusing to users and very hard to debug. We hit this in our code when running a Windows service and stderr was closed by Windows causing ConsoleWriter to fail. Current behavior depends on the order in which you configured the writers. Instead MultiLevelWriter should aggregate errors and return a single error.

nmldiegues:
Fixes #281

Currently MultiLevelWriter would give up if any individual backing logger
failed. This could result in no logging ever happening if the bad logger
was set up in the first position.

As a motivating factor, this can happen in "normal" circumstances, such
as running as a Windows Service where stderr is not available and therefore
the ConsoleWriter will fail. If you happen to set it as the first logger,
then no logging ever takes place. To make matters worse, connecting a
debugger creates an stderr, so it made it a pretty daunting situation
to overcome.

The proposed solution is to go through all underlying loggers and return
the last error obtained, if any. In practice there isn't much being done
with those errors anyway, as the best way to address logging errors is
to hook a ErrorHandler, which will still be triggered despite this change.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -56,30 +56,30 @@ type multiLevelWriter struct {
 
 func (t multiLevelWriter) Write(p []byte) (n int, err error) {
 	for _, w := range t.writers {
-		n, err = w.Write(p)
-		if err != nil {
-			return
-		}
-		if n != len(p) {
-			err = io.ErrShortWrite
-			return
+		if _n, _err := w.Write(p); err == nil {
+			n = _n
+			if _err != nil {
+				err = _err
+			} else if _n != len(p) {
+				err = io.ErrShortWrite
+			}
 		}
 	}
-	return len(p), nil
+	return n, err
 }
 
 func (t multiLevelWriter) WriteLevel(l Level, p []byte) (n int, err error) {
 	for _, w := range t.writers {
-		n, err = w.WriteLevel(l, p)
-		if err != nil {
-			return
-		}
-		if n != len(p) {
-			err = io.ErrShortWrite
-			return
+		if _n, _err := w.WriteLevel(l, p); err == nil {
+			n = _n
+			if _err != nil {
+				err = _err
+			} else if _n != len(p) {
+				err = io.ErrShortWrite
+			}
 		}
 	}
-	return len(p), nil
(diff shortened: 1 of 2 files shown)


writer.go
```diff
@@ -55,30 +55,42 @@ type multiLevelWriter struct {
 }
 
 func (t multiLevelWriter) Write(p []byte) (n int, err error) {
+	var anyErr error
```
rs, line 58: It’s counter intuitive to use err declared in the prototype as a local variable and to create another one to be returned. Would be better to use _err to test individual errors and keep using err as the return value for instance.

nmldiegues: I've simplified the error variable usage, it was indeed messy, well noted.

writer.go
```diff
@@ -57,28 +57,30 @@ type multiLevelWriter struct {
...
-		if err != nil {
-			return
-		}
 		if n != len(p) {
 			err = io.ErrShortWrite
-			return
 		}
 	}
+
+	if err != nil {
```
rs, line 65: This won't work, as if the last writer does not have an error, err will be overwritten with `nil`.

writer.go
```diff
@@ -54,31 +54,43 @@ type multiLevelWriter struct {
 	writers []LevelWriter
 }
 
-func (t multiLevelWriter) Write(p []byte) (n int, err error) {
+func (t multiLevelWriter) Write(p []byte) (int, error) {
```
rs, line 57: Please use this construct:

```go
func (t multiLevelWriter) Write(p []byte) (n int, err error) {
	for _, w := range t.writers {
		if _n, _err := w.Write(p); err == nil {
			n = _n
			if _err != nil {
				err = _err
			} else if _n != len(p) {
				err = io.ErrShortWrite
			}
		}
	}
	return n, err
 }
```

PR #284, opened 2021-01-21
Add Strs function to massive assignment string into array of log

insomnius:
**Problem**

There are repetitions when you want to do something like this:

```
zerolog.Arr().Str("some").Str("unecessary").Str("repetition")
```

**Proposal**

To simplify this assignment, I propose that we should have this kind of thing

```
zerolog.Arr().Strs("more", "efficient", "string", "assigment")
```

It would be a great coding experience if we could to this.

Tell me about your opinion @rs , should we do this also to other array assignment function? In this PR I just add `Strs` because that is the function that I really need to improve my coding experience :see_no_evil:

### DIFF
--- a/array.go
+++ b/array.go
@@ -67,6 +67,14 @@ func (a *Array) Object(obj LogObjectMarshaler) *Array {
 	return a
 }
 
+// Strs append append the val as a string to the array with massive assignment
+func (a *Array) Strs(val ...string) *Array {
+	for _, v := range val {
+		a.Str(v)
+	}
+	return a
+}
+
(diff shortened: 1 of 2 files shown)


rs:
This will allocate a slice.

insomnius:
> This will allocate a slice.

Yes it will, so what's the downside?

rs:
The zero in zerolog means zero allocation :)

insomnius:
> The zero in zerolog means zero allocation :)

Ah i see, then i will close this PR.

Issue #293, opened 2021-02-25
Get value from header or create with RequestIDHandler


artefactop:
Hi,

I'm trying to understand how `RequestIDHandler` works. I configured the HTTP handler this way (got from https://github.com/rs/zerolog/blob/master/hlog/hlog.go#L150 and changed a bit):

```go
func TestRequestIDFromHeaderHandler(t *testing.T) {
	out := &bytes.Buffer{}
	r := &http.Request{
		Header: http.Header{
			"Referer": []string{"http://foo.com/bar"},
			"X-Request-Id": []string{"514bbe5bb5251c92bd07a9846f4a1ab6"},
		},
	}
	h := RequestIDHandler("id", "X-Request-Id")(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		id, ok := IDFromRequest(r)
		if !ok {
			t.Fatal("Missing id in request")
		}
		if want, got := "514bbe5bb5251c92bd07a9846f4a1ab6", w.Header().Get("X-Request-Id"); got != want {
			t.Errorf("Invalid Request-Id header, got: %s, want: %s", got, want)
		}
		l := FromRequest(r)
		l.Log().Msg("")
		if want, got := fmt.Sprintf(`{"id":"%s"}`+"\n", id), decodeIfBinary(out); want != got {
			t.Errorf("Invalid log output, got: %s, want: %s", got, want)
		}
	}))
	h = NewHandler(zerolog.New(out))(h)
	h.ServeHTTP(httptest.NewRecorder(), r)
}
```

So if the request has already a header with that value, is ignored. But I see that the handler try to get the `idKey` from the context.

The problem I want to resolve if to create a new request ID only if it is not coming on the HTTP request.

Who is the responsible to set that value?

gilcrest:
Hi @artefactop  - you've set the X-Request-Id in the request header, but it needs to be set in the request context. I believe RequestIDHandler checks the request Context using the unexported `idKey` struct. If it does not find a value, it then adds one using @rs' `xid` library. The `headerName` function parameter that you're setting as `X-Request-Id` is used to add the request ID to the response headers and this field allows you to name that response header.

You could add logic in your handler chain to only call this handler if you don't have a value in that header... Sorry to jump in, hope I'm understanding what you're looking for!

artefactop:
Thank you, @gilcrest,

I do not understand why the following lines exists https://github.com/rs/zerolog/blob/master/hlog/hlog.go#L153
```go
[...]

PR #296, opened 2021-03-01
Allow setting context using idKey. Assist for issue #293.

gilcrest:
This is in reference to Issue #293. I actually have a separate PR ready as well, because it wasn't as simple as I thought it would be after digging into the code and testing. `RequestIDHandler` is tightly coupled to xid, so I did not want to change that. 

If @artefactop or others wants to be able to set the RequestID to the context prior to calling `RequestIDHandler`, it is possible with this new function. The only caveat is that the ID being set must be an `xid.ID`.

I have a whole separate PR with a CustomIDHandler and associated type, etc. to allow setting an arbitrary string as the Request ID. I did not want to do as one PR since you may not like the second one.

Let me know if you're ok with this as is.

Thanks!


rs:
Please remove go 7/8 from travis config so the test pass

gilcrest (commented):
Removed 1.7 and 1.8, added 1.16

PR #299, opened 2021-03-12
Fix: Ensure event.Stack() is called before event.Err()

krak3n:
Based from prior work on #179 by @larsla.

The author of the original PR has not addressed the change requests and since I just ran into this bug I thought I would make the changes and open a new PR to resolve the issue.

This fixes an issue where `zerolog.New().With().Stack().Logger()` would not add a stack trace since the hooks are applied after `Err()` is called.

### DIFF
--- a/context.go
+++ b/context.go
@@ -406,17 +406,10 @@ func (c Context) CallerWithSkipFrameCount(skipFrameCount int) Context {
 	return c
 }
 
-type stackTraceHook struct{}
-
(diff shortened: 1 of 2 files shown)


rs:
This won't do what you think. The stack trace will be captured at the moment of the context creation (once per logger) instead of the moment of the log event sending.

krak3n:
Ah I see, thanks @rs. I'll close this. Would you be point me in the right direction for a better way of solving this?

PR #301, opened 2021-03-18
Initial Earthfile for running tests

alexcb:
This PR is an example of how one can use [earthly](https://github.com/earthly/earthly) to perform integration tests both in github actions and locally.

# Running tests

### Docker

visit https://docs.docker.com/get-docker/ for installation instructions

### Earthly

Mac users can run:

    brew install earthly

and linux users can run:

    sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/latest/download/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly && /usr/local/bin/earthly bootstrap'

For additional information visit https://earthly.dev/get-earthly or https://github.com/earthly/earthly

## Testing

To run the tests locally for go 1.16, in the root of the repo run:

    earthly +test

To specific a different version of go (e.g. 1.14), run:

    earthly --build-arg GO_VERSION=1.14 +test

To test against all versions between 1.12 and 1.16, run:

    earthly +all

Alternatively you can run these tests even without checking out the git repo via:

    earthly github.com/alexcb/zerolog+test

### Interactive testing
If you run into a command that fails, you can include the `--interactive` flag which will drop you into a shell within the container that's running the command that failed.

For instance, I added a linter which was showing there's some potential code syntax issues and is returning an error exit code, so you could run it via an interactive shell with:

    earthly --interactive github.com/alexcb/zerolog+lint

Signed-off-by: Alex Couture-Beil <alex@earthly.dev>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,4 +26,4 @@ jobs:
       - name: Download released earthly
         run: "sudo /bin/sh -c 'wget https://github.com/earthly/earthly/releases/download/v0.5.7/earthly-linux-amd64 -O /usr/local/bin/earthly && chmod +x /usr/local/bin/earthly'"
       - name: test zerolog with go ${{ matrix.go_version }}
-        run: earthly --build-arg GO_VERSION=${{ matrix.go_version }} --use-inline-cache +test
+        run: earthly --build-arg GO_VERSION=${{ matrix.go_version }} --ci +test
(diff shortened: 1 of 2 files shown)


alexcb:
my forked version has an example of this passing via github actions: https://github.com/alexcb/zerolog/runs/2135672606

PR #306, opened 2021-04-07
Fix Typo

redhajuanda:


gilcrest:
@rs I actually think the original read better for this particular sentence? No big deal at all, but I think if you're going to change this, then it should read "Note that zerolog does not de-duplicate fields" instead of the current "Note that zerolog does not de-duplication of fields." IMHO the original sentence seemed correct and had the right intention.

rs:
You're right, I merged too fast. I will revert. Thanks.

PR #307, opened 2021-04-12
add Disabled to String and ParseLevel; add tests

gilcrest:
hi @rs - I've been working with ParseLevel recently and was wondering about `Disabled` vs. `NoLevel` - there wasn't any tests for the `Level.String` method and the `ParseLevel` function to help me understand the behavior, so I created some. I added Disabled as well to the `String()` switch statement and the `ParseLevel` case statement. I'm not sure if it was intentionally skipped for some reason? 

Thanks as always for your help!

Dan

### DIFF
--- a/log.go
+++ b/log.go
@@ -147,6 +147,8 @@ func (l Level) String() string {
 		return "fatal"
 	case PanicLevel:
 		return "panic"
+	case Disabled:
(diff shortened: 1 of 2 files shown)


gilcrest:
Thanks!

Issue #243, opened 2020-06-23
log.Logger.Write() writes wrong "caller" information (wrong frame skip count)


adw1n:
Many third party libraries let you configure logging by accepting an `io.Writer` object. To take advantage of full zerolog capabilities `Logger::Write()` function should adjust frame skip count in order to log "caller" information correctly.

example.go
```
package main

import (
	"github.com/rs/zerolog"
	"github.com/rs/zerolog/log"
)

func main() {
	log.With().CallerWithSkipFrameCount(zerolog.CallerSkipFrameCount + 1).Logger().Write([]byte("test"))
	log.Logger = log.With().Caller().Logger()
	log.Logger.Write([]byte("test"))
	log.Info().Msg("test")
}
```

output:
```
{"time":"2020-06-23T14:25:15+02:00","caller":"/home/scott/example.go:9","message":"test"}
{"time":"2020-06-23T14:25:15+02:00","caller":"/home/scott/go/pkg/mod/github.com/rs/zerolog@v1.19.0/log.go:407","message":"test"}
{"level":"info","time":"2020-06-23T14:25:15+02:00","caller":"/home/scott/example.go:12","message":"test"}
```

I expected
```
{"time":"2020-06-23T14:25:15+02:00","caller":"/home/scott/go/pkg/mod/github.com/rs/zerolog@v1.19.0/log.go:407","message":"test"}
```
to be
```
{"time":"2020-06-23T14:25:15+02:00","caller":"/home/scott/example.go:10","message":"test"}
```

This might be related to https://github.com/rs/zerolog/issues/98 depending on how you decide to fix it.

zelch:
I'm not sure if you're still waiting on this one @adw1n , but if you are I'd appreciate it if you could verify that my pull request really fixes it for your case as well.

PR #315, opened 2021-05-06
Fix handling of printing caller with Write, Print, and Printf.

zelch:
This fixes https://github.com/rs/zerolog/issues/243, as well as the identical issue around Print and Printf calls in both zerolog and zerolog/log.

With these changes, we now get correct caller data in all those cases.

### DIFF
--- a/event.go
+++ b/event.go
@@ -20,12 +20,13 @@ var eventPool = &sync.Pool{
 // Event represents a log event. It is instanced by one of the level method of
 // Logger and finalized by the Msg or Msgf method.
 type Event struct {
-	buf   []byte
-	w     LevelWriter
-	level Level
-	done  func(msg string)
-	stack bool   // enable error stack trace
-	ch    []Hook // hooks from context
+	buf                  []byte
+	w                    LevelWriter
+	level                Level
+	done                 func(msg string)
+	stack                bool   // enable error stack trace
+	ch                   []Hook // hooks from context
+	callerSkipFrameCount int    // The number of additional frames to skip when printing the caller.
 }
 
 func putEvent(e *Event) {
@@ -62,6 +63,7 @@ func newEvent(w LevelWriter, level Level) *Event {
 	e.w = w
 	e.level = level
 	e.stack = false
+	e.callerSkipFrameCount = 0
 	return e
 }
 
@@ -685,6 +687,13 @@ func (e *Event) Interface(key string, i interface{}) *Event {
 	return e
 }
 
+// CallerSkipFrame instructs any future Caller calls to skip the specified number of frames.
+// This includes those added via hooks from the context.
+func (e *Event) CallerSkipFrame(skip int) *Event {
+	e.callerSkipFrameCount++
+	return e
+}
+
 // Caller adds the file:line of the caller with the zerolog.CallerFieldName key.
 // The argument skip is the number of stack frames to ascend
 // Skip If not passed, use the global variable CallerSkipFrameCount
@@ -700,7 +709,7 @@ func (e *Event) caller(skip int) *Event {
 	if e == nil {
 		return e
 	}
-	_, file, line, ok := runtime.Caller(skip)
+	_, file, line, ok := runtime.Caller(skip + e.callerSkipFrameCount)
 	if !ok {
 		return e
 	}
--- a/event.go
+++ b/event.go
@@ -690,7 +690,7 @@ func (e *Event) Interface(key string, i interface{}) *Event {
 // CallerSkipFrame instructs any future Caller calls to skip the specified number of frames.
 // This includes those added via hooks from the context.
 func (e *Event) CallerSkipFrame(skip int) *Event {
-	e.skipFrame++
+	e.skipFrame += skip
 	return e
 }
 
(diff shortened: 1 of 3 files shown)


rs:
This is not thread safe.

zelch:
Indeed, it's not even safe for a single OS thread but multiple goroutines, I'm a bit annoyed with myself for failing to even think about the concurrent case.

I think I see how to do it more correctly, but the initial thought on that is uglier than I'd like.

I'll follow up in a few days when I've had time to play with a couple of different options.

rs:
If write is called concurrently (it's often the case with a logger), the counter will be arbitrary incremented. You can't write a global variable from a logger method. The best way to fix that would be to add a method on event to customized a skip frame delta for the message.

zelch:
I have just such a change working, but it will likely be tomorrow or the day after that I have the time to finalize it.

zelch:
Alright, the rework in the force-push should be significantly better in regards to thread (and basic concurrency) safety.

Obvious, if multiple goroutines are trying to use the same Event, we have problems, but I'm pretty sure that we have many other problems in that case as well.

log.go
```diff
@@ -392,6 +392,7 @@ func (l *Logger) Log() *Event {
 // Arguments are handled in the manner of fmt.Print.
 func (l *Logger) Print(v ...interface{}) {
 	if e := l.Debug(); e.Enabled() {
+		e.CallerSkipFrame(1)
```
rs, line 395: Please inline it the line below.

log.go
```diff
@@ -400,6 +401,7 @@ func (l *Logger) Print(v ...interface{}) {
 // Arguments are handled in the manner of fmt.Printf.
 func (l *Logger) Printf(format string, v ...interface{}) {
 	if e := l.Debug(); e.Enabled() {
+		e.CallerSkipFrame(1)
```
rs, line 404: Please inline it the line below.

log.go
```diff
@@ -412,6 +414,7 @@ func (l Logger) Write(p []byte) (n int, err error) {
 		// Trim CR added by stdlog.
 		p = p[0 : n-1]
 	}
+	e.CallerSkipFrame(1)
```
rs, line 417: `e` is not declared, does this compile?

event.go
```diff
@@ -20,12 +20,13 @@ var eventPool = &sync.Pool{
...
-	done  func(msg string)
-	stack bool   // enable error stack trace
-	ch    []Hook // hooks from context
+	buf                  []byte
+	w                    LevelWriter
+	level                Level
+	done                 func(msg string)
+	stack                bool   // enable error stack trace
+	ch                   []Hook // hooks from context
+	callerSkipFrameCount int    // The number of additional frames to skip when printing the caller.
```
rs, line 29: That's very long, what about `skipFrame`?

rs (changes requested):

log.go
zelch: It does not compile.  I blame my brain, as unhelpful as that is. :)

log.go
zelch: Done. 3326547548fb5fe2a8e6d4db82a7f692a17d6073

log.go
zelch: Done. 3326547548fb5fe2a8e6d4db82a7f692a17d6073

log.go
zelch: Fixed. 99c9830747e5fbe673073f2600e78ba9758cf0df

event.go
zelch: Sounds good.  Done.  3d15cd96b297c50fb0e980abbf94883c597572d7

event.go
```diff
@@ -685,6 +687,13 @@ func (e *Event) Interface(key string, i interface{}) *Event {
 	return e
 }
 
+// CallerSkipFrame instructs any future Caller calls to skip the specified number of frames.
+// This includes those added via hooks from the context.
+func (e *Event) CallerSkipFrame(skip int) *Event {
+	e.skipFrame++
```
rs, line 693: Should be `e.skipFrame += skip`

log.go
```diff
@@ -392,6 +392,7 @@ func (l *Logger) Log() *Event {
 // Arguments are handled in the manner of fmt.Print.
 func (l *Logger) Print(v ...interface{}) {
 	if e := l.Debug(); e.Enabled() {
+		e.skipFrame++
```
rs, line 395: When I said inline, I meant use the method on the same line as above, not actually inlining the function. Sorry for the confusion.

rs (changes requested):

log.go
zelch: That makes sense, I'll fix that up.

Would you object to putting it inline with the Msg call instead of the conditional line?

event.go
zelch: bdfa5fe2015622e12817a668b461f15fddfa3f9a

Zenithar:
For any indtermined reason, I have `e` in 

```
// CallerSkipFrame instructs any future Caller calls to skip the specified number of frames.
// This includes those added via hooks from the context.
func (e *Event) CallerSkipFrame(skip int) *Event {
	if e == nil {
		return e
	}
	e.skipFrame += skip
	return e
}
```

So I added a nil check first.

zelch:
> For any undetermined reason, I have `e` nil in

That's reasonable, I was going to say that it seems somewhat insane to try and do anything with a nil Event, except most of the other functions that work on Event check for nil, so we should as well.

PR #318, opened 2021-05-17
Add ability to customize internal json marshaler

### LINKED ISSUE #317: Feature: customizd interface marshaler
Currently, when using `logger.Debug().Interface("request_body", reqBody).Msg("request sent")` to log the request detail with customized types, zerolog first assert if the "reqBody" is a `LogObjectMarshaler` or not, if not, it tries the std json marshaler to serialize the request body.

As you can see in the code snippet. It's hard coded to `json.Marshal`.
https://github.com/rs/zerolog/blob/3c3b4a354e411c2af7f7ac649696cc5a590f2381/internal/json/types.go#L363-L371

However, for performance and convenient, mostly we do not use the standard json package for json encode/decode now.
I hope zerolog could have a package level `InterfaceMarshaler` variable, and use it to marshal object when needed.
It has the same signature with `json.Marshal`, so everyone can replace the internal reflection based marshaler at runtime without forking zerolog.
```go
type InterfaceMarshaler func(obj interface{}) (data []byte, err error)
```

PS: I am using [jsoniter](https://github.com/json-iterator/go) and have registered many customized encoders/decoders for customized types. That's why I hope zerolog can benefit from jsoniter without additional works.

povsister:
Added a package level variable `InterfaceMarshalFunc`. It's used to marshal interface to JSON encoded byte slice, mostly when event.Interface("key", v) is called.

This enables user to replace the internal json marshaler at runtime, be noticed that using it improperly will cause potential data race.
e.g: Changing `InterfaceMarshalFunc` while lots of events are being processed.

Fixes #317

### DIFF
--- a/internal/cbor/base.go
+++ b/internal/cbor/base.go
@@ -1,5 +1,13 @@
 package cbor
 
+// JSONMarshalFunc is used to marshal interface to JSON encoded byte slice.
+// Making it package level instead of embedded in Encoder brings
+// some extra efforts at importing, but avoids value copy when the functions
+// of Encoder being invoked.
+// DO REMEMBER to set this variable at importing, or
+// you might get a nil pointer dereference panic at runtime.
+var JSONMarshalFunc func(v interface{}) ([]byte, error)
+
 type Encoder struct{}
 
 // AppendKey adds a key (string) to the binary encoded log message
--- a/encoder_cbor.go
+++ b/encoder_cbor.go
@@ -14,6 +14,13 @@ var (
 	enc = cbor.Encoder{}
 )
 
+func init() {
+	// using closure to reflect the changes at runtime.
+	cbor.JSONMarshalFunc = func(v interface{}) ([]byte, error) {
+		return InterfaceMarshalFunc(v)
+	}
+}
+
 func appendJSON(dst []byte, j []byte) []byte {
 	return cbor.AppendEmbeddedJSON(dst, j)
 }
--- a/encoder_json.go
+++ b/encoder_json.go
@@ -15,6 +15,13 @@ var (
 	enc = json.Encoder{}
 )
 
+func init() {
+	// using closure to reflect the changes at runtime.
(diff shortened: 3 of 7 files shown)


encoder.go
```diff
@@ -1,10 +1,23 @@
...
 	"net"
 	"time"
 )
 
+var (
+	// JSONMarshaler is used to marshal interface
+	// to JSON encoded byte slice when an interface
+	// is passed for logging.
+	// Default: "encoding/json.Marshal"
+	JSONMarshaler InterfaceMarshaler = json.Marshal
```
rs, line 14: I would rather have this variable named `InterfaceMarshalFunc` and remove `InterfaceMarshaler` type that is not very useful. Please update the comment so it does not wrongly implies this is an interface.

Also, please move this to the `globals.go` file.

internal/cbor/base.go
```diff
@@ -1,5 +1,13 @@
 package cbor
 
+// JSONMarshaler knows how to marshal an interface to JSON encoded byte slice.
+// Making it package level instead of embedded in Encoder brings
+// some extra efforts at importing, but avoids value copy when
+// the functions of Encoder being invoked.
+// DO REMEMBER to set this variable at importing, or
+// you might get a nil pointer dereference panic at runtime.
+var JSONMarshaler func(v interface{}) ([]byte, error)
```
rs, line 9: Why nor init this as `var JSONMarshFunc = json.Marshal`?

rs (changes requested):

internal/cbor/base.go
povsister: Because it's an internal package, designed for non-independent usage. I'd like to keep the setting explicitly to avoid someone happens to left out the configuration. Also I think it makes the configuration clear and global once, no confusion will be made with the default marshal func.

encoder.go
povsister: I agree. Will push another fix till all coversation resolved.

povsister:
@rs Updated. PTAL

rs (approved):

PR #321, opened 2021-05-19
Add Func log method

marwatk:
This is adds the `Func` log method to log using an anonymous function
only if the level is currently enabled.

The use case is for when you don't own an object and therefore can't
create your own marshaller but need to do some translation prior to
logging.

For example, this:

```go
msg := log.Debug()
if msg.Enabled() {
  msg.Str("complicated_thing", makeBinaryThingLoggable(thing))
}
msg.Msg("Sending complicated thing")
```

Turns into this:

```go
log.Debug().
  Func(func(e *Event) { e.Str("complicated_thing", makeBinaryThingLoggable(thing)) }).
  Msg("Sending complicated thing")
```

I considered a `StrFunc` instead which would be simpler but less flexible. I'm open to either and the possibility I'm missing a much simpler way to accomplish the same thing.

### DIFF
--- a/event.go
+++ b/event.go
@@ -209,6 +209,14 @@ func (e *Event) Object(key string, obj LogObjectMarshaler) *Event {
 	return e
 }
 
+// Func allows an anonymous func to run only if the event is enabled.
+func (e *Event) Func(f func(e *Event)) *Event {
+	if e != nil && e.Enabled() {
+		f(e)
+	}
+	return e
+}
+
 // EmbedObject marshals an object that implement the LogObjectMarshaler interface.
 func (e *Event) EmbedObject(obj LogObjectMarshaler) *Event {
 	if e == nil {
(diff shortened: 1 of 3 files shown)


event.go
```diff
@@ -209,6 +209,15 @@ func (e *Event) Object(key string, obj LogObjectMarshaler) *Event {
 	return e
 }
 
+// Func allows an anonymous func to run only if the event is enabled.
+func (e *Event) Func(f func(e *Event)) *Event {
+	if e == nil {
+		return e
+	}
+	f(e)
```
rs, line 217: You still have to check if enabled as an event can be non nil with the level `Disabled`.

rs (changes requested):

event.go
marwatk: Fixed.

rs:
Would be nice to add a usage example in the readme next to the other method.

marwatk:
> Would be nice to add a usage example in the readme next to the other method.

@rs, I'm not 100% sure where you mean. No sections in the readme are jumping out at me as a logical place to put it. Do you have a specific section in mind?

rs (approved):

rs:
You're right, I thought the `Enable` example was in its own section.

PR #336, opened 2021-07-26
Change global to time.RFC3339Nano. Fixes #213

### LINKED ISSUE #213: ConsoleWriter TimeFormat
I created a new logger like this:
```go
l := zerolog.New(zerolog.ConsoleWriter{
	Out:        os.Stderr,
	TimeFormat: "2006-01-02T15:04:05.999Z07:00",
}).With().Timestamp().Logger()
```
However, the time logged by it looked like this:
`2020-02-05T10:27:38+01:00 ERR server handle new connection`

Looking at the code and how the console writer formats the time, it turned out I had to globally set the time field format to something that is formatting AT LEAST as precise, as the format on the ConsoleWriter that I want. This means setting:
`zerolog.TimeFieldFormat = time.RFC3339Nano`
yielded the correct time output:
`2020-02-05T10:27:38.921+01:00 ERR server handle new connection:`

Is this how it is intended? Setting the global time field of course affects all my loggers...

kishaningithub:
Context #213

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -83,7 +83,7 @@ var (
 	// TimeFieldFormat defines the time format of the Time field type. If set to
 	// TimeFormatUnix, TimeFormatUnixMs or TimeFormatUnixMicro, the time is formatted as an UNIX
 	// timestamp as integer.
-	TimeFieldFormat = time.RFC3339
+	TimeFieldFormat = time.RFC3339Nano
 
 	// TimestampFunc defines the function called to generate a timestamp.
 	TimestampFunc = time.Now


rs:
Changing a default setting could break existing users of the library not expecting the time format to change.

kishaningithub:
@rs Agreed.

Do you have any elegant way to address #213 that does not involve consumer breakages

Would be happy to contribute a PR for the same ☺️

mitar:
As I wrote in #213, maybe an entry in the README should explain the interaction between `zerolog.TimeFieldFormat` and console writer instead of just changing the default.

PR #347, opened 2021-08-19
allow RemoteAddrHandler to include addresses which have no port

Vivalldi:
Many HTTP servers end up being placed behind reverse proxies. As a result, many devs use middleware to map reverse-proxy headers to the request RemoteAddr. Many of these headers don't include ports, causing an otherwise valid RemoteAddrHandler to not set the field.﻿

### DIFF
--- a/hlog/hlog.go
+++ b/hlog/hlog.go
@@ -3,7 +3,6 @@ package hlog
 
 import (
 	"context"
-	"net"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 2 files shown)


rs:
We should probably not remove the port anyway. I would accept a PR with plain remote addr copied to the field.

Vivalldi:
Just for clarification, are you suggesting

A. Update `RemoteAddrHandler` to only use the `r.RemoteAddr` in it's pure state (unless empty, then don't add)
```
if r.RemoteAddr != "" {
    // Add to log
}

B. Create a new Handler that does the above, but leaves `RemoteAddrHandler` intact

rs:
I mean A. This is a change of behavior, but thinking about it, current behavior is misleading.

Vivalldi:
Updated! Would you like to keep or remove the additional tests? (They don't really assert anything special anymore due to the change in logic)

rs:
You can remove the non relevant ones.

Vivalldi:
Done! Thanks for the feedback

PR #352, opened 2021-08-31
Fields helper accepts both map and slice

hn8:
Useful for logging wrappers like [zerologr ](https://github.com/hn8/zerologr) to reuse the same code for Fields helper (which requires map). No extra allocation for variadic arguments or slice of key/value pairs. It avoids stack copy overhead from variadic arguments (as in Zap ...Fields)

### DIFF
--- a/benchmark_test.go
+++ b/benchmark_test.go
@@ -148,16 +148,16 @@ func BenchmarkLogFieldType(b *testing.B) {
 		{"a", "a", 0},
 	}
 	objects := []obj{
-		obj{"a", "a", 0},
-		obj{"a", "a", 0},
-		obj{"a", "a", 0},
-		obj{"a", "a", 0},
-		obj{"a", "a", 0},
-		obj{"a", "a", 0},
-		obj{"a", "a", 0},
-		obj{"a", "a", 0},
-		obj{"a", "a", 0},
-		obj{"a", "a", 0},
+		{"a", "a", 0},
+		{"a", "a", 0},
+		{"a", "a", 0},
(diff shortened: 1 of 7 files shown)


rs:
I'm not a big fan of the name of the method. Ideally it should be `Fields` but it is taken by the map form. I also wonder if a variadic argument wouldn't be more friendly for this method (not sure about the allocation difference).

What I have in mind (but convinced by any of them):

* `Fields(interface)` that would handle both map and array forms via type inference. I generally don't like type-less APIs, but here it is more a convenience method so perhaps it's ok
* `Fields(interface…)` same as above but either it's a single map arg or a variadic key, value… It might be very confusing.
* `FieldList([]interface)` isn't much more convincing than `KeyValues`

Ideas are welcome.

hn8:
@rs Pushed the change to go with first option Fields(interface). It is less confusing than variadic version, also has less overhead. Tried to optimize logr Zap wrapper before, learned that passing arguments with variadic spread down to call stack does hurt its performance. Does it look good for you?

rs:
Looks good. Could you please add some tests and an example?

hn8:
@rs 👍

hn8:
@rs Would you please create a new git tag (for go mod)? Thanks!

rs:
done

tonglil:
Just wondering, is there any way Fields() can detect errors and call `Err(err)` to print the stacktrace for https://github.com/rs/zerolog#error-logging-with-stacktrace?

Example: https://go.dev/play/p/ulHc0WIqlC0

PR #354, opened 2021-09-04
zerolog.Context now knows how to LoggerRef()

### LINKED ISSUE #116: Chaining WithContext call is broken
#81 broke clients that chain `WithContext` calls. According to the docs, I expect the following program to compile (lifted from #80):

```
package main

import (
	"context"
	"github.com/rs/zerolog"
	"os"
)

	child(log.WithContext(ctx))
}

}

func main() {
	process(ctx, 1)

	process(ctx, 2)

}
```

but it fails: 
```
# command-line-arguments
cmd/main.go:19:31: cannot call pointer method on zerolog.New(os.Stdout)
cmd/main.go:19:31: cannot take the address of zerolog.New(os.Stdout)
```

Reason being that variables in go are addressable but return values from functions are not.

The current workaround is this: 
```
package main

import (
	"context"
	"github.com/rs/zerolog"
	"os"
)

	child(log.WithContext(ctx))
}

}

func main() {
	l := zerolog.New(os.Stdout)
	ctx := l.WithContext(context.Background())
	process(ctx, 1)

	process(ctx, 2)

}
```

Output:
```
```

rsrchboy:
This should be simple enough -- same as `Logger()`, but with a different return, useful for chaining with `WithContext()`.

I've included a simple test case, just to demonstrate this addresses the issue in https://github.com/rs/zerolog/issues/116

Fixes #116

### DIFF
--- a/context.go
+++ b/context.go
@@ -18,6 +18,11 @@ func (c Context) Logger() Logger {
 	return c.l
 }
 
+// LoggerRef is the same as Logger(), but returns a *Logger.
+func (c Context) LoggerRef() *Logger {
+	return &c.l
+}
+
 // Fields is a helper function to use a map to set fields using type assertion.
 func (c Context) Fields(fields map[string]interface{}) Context {
 	c.l.context = appendFields(c.l.context, fields)
--- a/context_test.go
+++ b/context_test.go
@@ -1,0 +1,24 @@
+package zerolog_test
+
+import (
+	"bytes"
+	"context"
+	"testing"
+
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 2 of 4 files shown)


rs:
This is confusing and is bloating the lib API.

rsrchboy:
So, in #116 you invited a fix / PR to address the issue; this is the approach that was discussed.

Do you have an alternate approach to suggest, or should #116 be closed as wontfix?

rs:
The discussed solution was to make WithContext a non pointer method. See https://github.com/rs/zerolog/issues/116#issuecomment-456327854

rsrchboy:
Ah.  I was going off of the prior comment https://github.com/rs/zerolog/issues/116#issuecomment-456295130, and looking for a minimal "won't break anything" solution.

PR #358, opened 2021-09-13
Allow to exclude fields from console output

ImSingee:


ImSingee:
@rs Sorry, but it seems that nothing is merged...

PR #360, opened 2021-09-17
Add Stringers support

ramasauskas:


rs:
Please add the cbor version too.

ramasauskas:
@rs

rs:
Why not using the new `enc.AppendStringer` in the existing `Event.Stringer`?

ramasauskas:
@rs event now uses encoder's `AppendStringer`.

PR #364, opened 2021-09-25
Update zerologr url

hn8:
Transferred to go-logr


hn8:
@rs Finally it is hosted by logr now, thanks!

Issue #329, opened 2021-07-08
Support logger-level ErrorMarshalFunc


Deiz:
I have a use case around dispatching errors with additional context, e.g. request IDs.

Currently, I'm accumulating context in HTTP middleware via the usual `With().Str("foo", "bar")` mechanism, which goes into the logger's `context []byte`.

I dispatch errors via a custom `ErrorMarshalFunc`, however the signature for that function, `func(err error) interface{}`, prevents obtaining any context from the logger or the event, which is fine.

What I'm proposing is to add an additional field, `errorMarshalFunc func(err error) interface{}` onto Logger, and an additional method, `func (l Logger) ErrorMarshalFunc(func(err error) interface{}) Logger` akin to Logger's Output method.

 That would support a closure-based flow like so:

```go
// Middleware
attributes := map[string]string{"foo": "bar"}

baseLogger := hlog.FromRequest(r)

errorMarshalFunc := func(err error) interface{} {
    // attributes captured from outer scope; this isn't possible with the global ErrorMarshalFunc
    externalerrors.Submit(err, attributes)

    return err
}

logger := baseLogger.With().Str("foo", "bar").Logger().ErrorMarshalFunc(errorMarshalFunc)

// Much later, in application code. Because of the request context's logger's embedded errorMarshalFunc, it retains the
// request-scoped metadata.
hlog.FromRequest(r).Err(err).Msg("Whoops")
```

This is probably a somewhat obtuse use case, but I'll be doing this work regardless and would like to see if there's consensus to merge this here.

rs:
I have the feeling you are trying to use this logging library as an event dispatcher to do other things. How is `externalerrors.Submit` related to the JSON logging?

Deiz:
Cheers for the quick response - You're spot on. For full real-world context:
- We use zerolog everywhere, with `log.Err(err)...` being muscle memory for our engineers
- We're looking to roll out SDK-driven error reporting, rather than parsing e.g. stringified `runtime/stack` output from zerolog's JSON; that typically looks like [rollbar.Error(err)](https://pkg.go.dev/github.com/rollbar/rollbar-go#Error) or [sentry.CaptureException(err)](https://pkg.go.dev/github.com/getsentry/sentry-go#CaptureException)

[...]

PR #365, opened 2021-09-25
feat: implement an optional ErrorWriter interface

Deiz:
@rs no idea if you'll want to merge this - this is something I'll be adopting internally so that errors logged via zerolog at warn level or above can be enriched with request context, then pushed into an external error aggregator; Sentry in my case.

This builds on the idea proposed in https://github.com/rs/zerolog/issues/329

A few design notes:
- The reason I went for an optional interface on the writer is because
  errors need level context, generally. In my use case, I don't want to
  send errors below "warn" level to an external error aggregator.
- Intended use of this requires an ErrorWriter implementation, which
  should be injected into the logger on a per-request basis such that
  all logged errors are enriched with the same request context.
- The []error embedded within the Logger, Event, and Array structs is
  there to guarantee Discard() behaves correctly, and also that e.g. an
  Array that is built up can benefit from the ErrorWriter that may be
  embedded within the event it will eventually be attached to.

### DIFF
--- a/array.go
+++ b/array.go
@@ -17,7 +17,8 @@ var arrayPool = &sync.Pool{
 // Array is used to prepopulate an array of items
 // which can be re-used to add to log messages.
 type Array struct {
-	buf []byte
+	buf  []byte
+	errs []error
 }
 
 func putArray(a *Array) {
@@ -38,6 +39,7 @@ func putArray(a *Array) {
 func Arr() *Array {
 	a := arrayPool.Get().(*Array)
 	a.buf = a.buf[:0]
+	a.errs = a.errs[:0]
 	return a
 }
 
@@ -93,6 +95,10 @@ func (a *Array) RawJSON(val []byte) *Array {
 
 // Err serializes and appends the err to the array.
 func (a *Array) Err(err error) *Array {
+	if err != nil {
+		a.errs = append(a.errs, err)
+	}
+
 	switch m := ErrorMarshalFunc(err).(type) {
 	case LogObjectMarshaler:
 		e := newEvent(nil, 0)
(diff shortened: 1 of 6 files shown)


Deiz:
@rs I'm curious if you'd be open to this or a similar change at this point.

I was in the middle of updating this branch (it's over two years old now 🙂) for internal usage and noticed that https://github.com/rs/zerolog/pull/559 solves _some_ of my use case, but ultimately the core issue that motivated this PR remains.

Chiefly, Zerolog doesn't offer a way to retain errors as `error` - they're immediately marshalled, just like everything else, which means losing some fidelity in custom error types, etc.

My question to you is whether you feel the use case of needing to do something special with errors when a Zerolog event is written rises to the level where you'd consider it worthy of being integrated into Zerolog itself.

rs:
I think this should not be part of a logging library, but part of your code that handle errors.

Deiz:
There are use cases like the following, where one might want to log, but otherwise ignore an error:
```go
for _, item := range items {
  result, err := fallibleOp(item)
  if err != nil {
    log.Err(err).Str("id", item.ID).Msg("Failed to perform operation")
    continue
  }

  ...
}
```

That's where we find it useful to ensure that we dispatch errors externally (to Sentry, in my company's case) as an operation that's coupled with actually writing the Zerolog event out.

In scenarios where errors are fatal and you bubble them up, I agree that having a standard error handler a) write to third-party integrations, b) log via Zerolog makes complete sense.

There's nothing stopping us from having something that looks like `func ReportError(ctx context.Context, event *zerolog.Event, message string, err error)` to accomplish this in scenarios like the above example, of course - it'd just be nice to have this in Zerolog's own ergonomics.

That said, seems like you're got a pretty strict separation of concerns in mind - so I suspect this'll remain a behaviour unique to my fork.

rs:
Yes I think the logger should be a destination and not a hub for all your app events. For instance if you decide to sample your logs, should your errors handlers be sampled as well? Probably not.

Issue #366, opened 2021-09-28
Default console formatter use same color for field and value on Err


AsgerNoer:
This is not anything critical and totally a quality of life fix, but the output of the `ConsoleWriter` when writing following code snippet will not color the error field name with the same color as all the other field names.

```go
log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})

if err := SomeFunc(); err != nil {
  log.Info().Err(err).Msg("log message")
}
```

### Function in question
https://github.com/rs/zerolog/blob/07d347e658dc44d677db54a67cd0e416d7c172da/console.go#L399-L403

PR #367, opened 2021-09-28
fix: error field name color

AsgerNoer:
This is a fix for #366 

Changed the color to reflect the other field names color


AsgerNoer:
@rs  a small change

Issue #368, opened 2021-09-28
Support for testing.T Logging


emcfarlane:
Feature request, like zaptest, to support `t.Log` for nicely capturing logs for testing.

rs:
You'll have to be more specific.

emcfarlane:
Ah sorry, I meant like the zap package: https://pkg.go.dev/go.uber.org/zap/zaptest . Which wraps the zap logger. It's a nice way to easily provider a logger that hooks into t.Log. Maybe there is a neat way to do this already that I missed?

rs:
Do you have a use-case example. I’m failing to see when this is useful.

emcfarlane:
For example if I create a test in a `t.Run(..)` like:
- TestWithZaptest `l := zaptest.NewLogger(t)`
- TestWithZerologLog `log`
- TestWithZerolog `out := &bytes.Buffer{}; l := zerolog.New(out)`

The output of the test cases is shown below, first without `-v` then with. It ties the log output very neatly to the testcase. Which helps when logging gets more complicated.

```
$ tmp go test .
--- FAIL: TestWithZaptest (0.00s)
    --- FAIL: TestWithZaptest/ok (0.00s)
        logger.go:130: 2021-09-29T11:25:08.766+0100     INFO    hello zaptest   {"name": "ok"}
        logger.go:130: 2021-09-29T11:25:08.766+0100     ERROR   zaptest error   {"ok": false}
{"level":"info","name":"ok","time":"2021-09-29T11:25:08+01:00","message":"hello zerolog"}
{"level":"error","ok":false,"time":"2021-09-29T11:25:08+01:00","message":"zerolog error"}
{"level":"info","name":"bad","time":"2021-09-29T11:25:08+01:00","message":"hello zerolog"}
--- FAIL: TestWithZerologLog (0.00s)
    --- FAIL: TestWithZerologLog/ok (0.00s)
--- FAIL: TestWithZerolog (0.00s)
    --- FAIL: TestWithZerolog/ok (0.00s)
        main_test.go:83: {"level":"info","name":"ok","message":"hello zerolog"}
            {"level":"error","ok":false,"message":"zerolog error"}
FAIL
FAIL    github.com/emcfarlane/tmp       0.126s
FAIL
$ tmp go test -v .
=== RUN   TestWithZaptest
=== RUN   TestWithZaptest/ok
[...]

PR #369, opened 2021-09-30
New TestWriter for logging to testing.TB

emcfarlane:
Implements a writer to wrap testing.TB for help in testing, https://github.com/rs/zerolog/issues/368.

The go testing log injects the callers filename and line number, which is incorrect for deeply nested functions. To get around this we allow to specify the frame depth and rewrite it. Not necessary but makes it pretty. Added an option for the ConsoleWriter to set the correct frame (although it's brittle to rewrites, not sure how to test better).

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -107,6 +107,7 @@ func MultiLevelWriter(writers ...io.Writer) LevelWriter {
 type TestingLog interface {
 	Log(args ...interface{})
 	Logf(format string, args ...interface{})
+	Helper()
 }
 
 // TestWriter is a writer that writes to testing.TB.
@@ -119,11 +120,13 @@ type TestWriter struct {
 
 // NewTestWriter creates a writer that logs to the testing.TB.
 func NewTestWriter(t TestingLog) TestWriter {
-	return TestWriter{T: t, Frame: 1}
+	return TestWriter{T: t}
 }
 
 // Write to testing.TB.
 func (t TestWriter) Write(p []byte) (n int, err error) {
+	t.T.Helper()
+
 	n = len(p)
 
 	// Strip trailing newline because t.Log always adds one.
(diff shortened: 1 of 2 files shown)


writer.go
```diff
@@ -96,3 +102,45 @@ func MultiLevelWriter(writers ...io.Writer) LevelWriter {
...
+// TestWriter is a writer that writes to testing.TB.
+type TestWriter struct {
+	TB testing.TB
+
+	// Frame skips caller frames to capture the original file and line numbers.
+	Frame int
+}
+
+// NewTestWriter creates a writer that logs to the testing.TB.
+func NewTestWriter(t testing.TB) TestWriter {
```
rs, line 115: Instead of importing `testing`, can we define a more restricted interface here?

writer.go
```diff
@@ -96,3 +102,45 @@ func MultiLevelWriter(writers ...io.Writer) LevelWriter {
...
+
+// Write to testing.TB.
+func (t TestWriter) Write(p []byte) (n int, err error) {
+	n = len(p)
+
+	// Strip trailing newline because t.Log always adds one.
+	p = bytes.TrimRight(p, "\n")
+
+	// Try to correct the log file and line number to the caller.
+	if t.Frame > 0 {
```
rs, line 127: Wouldn't `t.TB.Helper()` do that for you?

writer.go
```diff
@@ -96,3 +102,45 @@ func MultiLevelWriter(writers ...io.Writer) LevelWriter {
...
+		_, origFile, origLine, _ := runtime.Caller(0)
+		_, frameFile, frameLine, ok := runtime.Caller(t.Frame)
+		if ok {
+			erase := strings.Repeat("\b", len(path.Base(origFile))+len(strconv.Itoa(origLine))+3)
+			t.TB.Logf("%s%s:%d: %s", erase, path.Base(frameFile), frameLine, p)
+			return n, err
+		}
+	}
+	t.TB.Logf("%s", p)
```
rs, line 136: Why not `t.TB.Log(string(p))`?

rs (changes requested):

writer.go
emcfarlane: Helper moves the problem up one, but there’s too many function calls that don’t know about the ‘t.Helper()’. For example, the ConsoleLogger with helper will set the log prefix to: ‘buffer.go:293: ‘ as it calls bytes buffer WriteTo method on the writer.

writer.go
emcfarlane: Changed to `TestingLog` interface of the subset of `Log` methods.

writer.go
emcfarlane: Changed to calling `t.Helper` to have a better zero frame default.

writer.go
```diff
@@ -96,3 +102,54 @@ func MultiLevelWriter(writers ...io.Writer) LevelWriter {
...
+			return n, err
+		}
+	}
+	t.T.Log(string(p))
+
+	return n, err
+}
+
+// ConsoleTestWriter creates an option that correctly sets the file frame depth for testing.TB log.
+func ConsoleTestWriter(t testing.TB) func(w *ConsoleWriter) {
```
rs, line 151: This should be `TestingLog`.

Issue #128, opened 2019-01-29
Positioning context fields in ConsoleWriter


manuuncle:
As mentioned in #50, I understand that the ordering of context fields is not considered to improve performance. 

As suggested in the ticket, can we do positioning the context fields and values using `PartsOrder` in ConsoleWriter ?

PR #383, opened 2021-12-04
ConsoleWriter: remove fields written within "PartsOrder" so they are not duplicated in output log

### LINKED ISSUE #204: Provide a way to exclude fields from ConsoleWriter
For example, you can set up `ConsoleWriter` in such a way that custom fields are in `PartsOrder`, like this:

```
output := zerolog.ConsoleWriter{Out: os.Stderr}
output.PartsOrder = []string{
	zerolog.TimestampFieldName,
	zerolog.LevelFieldName,
	"node",
	zerolog.MessageFieldName,
}

logger := zerolog.New(output).With().Timestamp().Str("node", node).Logger()
return &logger
```

However, there's no way of excluding the custom field from then being duplicated in the `writeFields` call. In addition, there's no way to control the rendering of custom fields in `PartsOrder` for e.g. adding padding for alignment.

The feature request is thus:
- Provide a way to exclude fields in the `ConsoleWriter`, or automatically exclude fields from `writeFields` if the field is specified in `PartsOrder` (the latter makes the most sense to me).
- `ConsoleWriter.FormatFieldValue` should have the field name passed to it, so that the presentation of those custom fields can be modified correctly.

adamko147:
close #204 

once the field is written into Parts, remove it form `evt` so it's no longer written to fields

### DIFF
--- a/console.go
+++ b/console.go
@@ -109,6 +109,7 @@ func (w ConsoleWriter) Write(p []byte) (n int, err error) {
 
 	for _, p := range w.PartsOrder {
 		w.writePart(buf, evt, p)
+		delete(evt, p)
 	}
 
 	w.writeFields(evt, buf)


DanMan259 (approved):
Tested and completes the task intended

DanMan259:
Solves issues #446 #424 #128

adubkov:
Why this was reverted? It the issues is valid.

adubkov:
Ok. I guess I found the answer. There is a workaround for it by using a combination of `PartsOrder` and `FieldsExclude`:

```
    log.Logger = log.Output(zerolog.ConsoleWriter{
        Out:        os.Stderr,
        TimeFormat: time.RFC3339,
        PartsOrder: []string{
            zerolog.TimestampFieldName,
            "ctx",
            zerolog.LevelFieldName,
            zerolog.CallerFieldName,
            zerolog.MessageFieldName,
        },
        FieldsExclude: []string{"ctx"},
    })
```

PR #387, opened 2021-12-15
Updated the crypto dependancy to force it to use latest

chrisharrisonkiwi:
Updated the dependency golang.org/x/crypto to force it to use the latest version.
This will remove the 2021 [security concern](https://cwe.mitre.org/data/definitions/347.html) brought up by Mitre:
[CVE-2020-9283](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-9283)

Also reported by the team at [Snyk](https://security.snyk.io/):
https://security.snyk.io/vuln/SNYK-GOLANG-GOLANGORGXCRYPTOSSH-551923

PR #389, opened 2021-12-16
move the lint command to its own package

drakkan:
in this way we remove several dependencies from the actual library.

This is just an idea. The logging library does not depends on crypto anymore for example.

See #387


drakkan:
Thank you

Issue #392, opened 2022-01-04
log.Fatal() and log.Panic() are not guaranteed to exit


a-canya:
In my app I initialize the global log level with:

```go
level, err := zerolog.ParseLevel(logLevel)
if err != nil { ... }
zerolog.SetGlobalLevel(level)
```

and then I use the global logger to do all kinds of stuff. For instance:

```go
log.Fatal().Msg("something happened and app needs to end")
```

However, if the `logLevel` was something such as `"panic"` or `"disabled"`, the log.Fatal() will not end the program. What's even more annoying is that if `logLevel` is an empty string, then level will be NoLevel and nothing will be logged (and fatals and panics will not cause the program to exit). [(playground example)](https://go.dev/play/p/T4aniqI-j0l)

The workaround is quite obvious and easy if you are aware of this problem. But taking into account that the native log package has [equivalent functions](https://pkg.go.dev/log#Fatal) which are guaranted to exit, this seems confusing to me.

Also, while coding, if you really want the program to end, you may consider calling sys.Exit() rather than calling log.Fatal().

- It would be less confusing to me if log.Fatal() and log.Panic() would always call sys.Exit() / panic(), even if their logging level is disabled and nothing is logged, so that program flow does not depend on log level. Is the current behaviour a design choice? What do you think about changing this?

- If this is not changed, I suggest at least adding a warning in the docs, at the log.Fatal log.Panic, Logger.Fatal and Logger.Panic functions and methods.

rs:
You are right, in `newEvent`, done should be called before returning when `enabled` is false. Would you send a PR?

a-canya:
Hi @rs , thanks for the quick reply. I am not very familiar with the codebase but I'm taking a look.

If I understand your proposal and the code correctly, this would have weird consequences:

```go
zerolog.SetGlobalLevel("info")
log.Fatal() // does not exit because Msg() is not called (current behaviour)
```

```go
zerolog.SetGlobalLevel("disabled")
log.Fatal() // exits because we call done at newEvent()
```

rs:
That’s correct

a-canya:
I've been looking for a way to improve what I pointed out in the previous comment but it would require many changes. Another issue with this solution is that the message is lost in the case of panic.

Anyway, I guess this is a rather unusual corner case and keeping the os.Exit() / panic() call is an improvement.

PR #393, opened 2022-01-04
Call exit/panic in Fatal/Panic log events even if the logging is disabled for this event

a-canya:
Issue #392

### DIFF
--- a/log.go
+++ b/log.go
@@ -428,6 +428,9 @@ func (l Logger) Write(p []byte) (n int, err error) {
 func (l *Logger) newEvent(level Level, done func(string)) *Event {
 	enabled := l.should(level)
 	if !enabled {
+		if done != nil {
+			done("")
+		}
 		return nil
 	}
 	e := newEvent(l.w, level)


rs:
Looks good. Keep in mind the panic won't have the message either.

rs:
One way to workaround the issue of panic with no message would be to change the Event's msg method to always call `done` even if the event is `nil`. The `nil` check would have to be moved from `Msg`/`Msgf` to `msg`.

rs:
Ignore my previous message, you can't have a `nil` event and `e.done` set…

a-canya:
That's true... The only workaround I can think of is to make the event be non-nil when it is disabled, and instead have an extra bool field in the struct to indicate that the event is disabled. Then apart of nil checks we would also need to check that the event is not disabled. This means more changes and I'm not sure what other implicactions it might have.

rs:
True, I'm not a big fan of such large change. Let's see if this trips more people as it's a bit of an edge case.

Issue #290, opened 2021-02-08
Dynamic Hooks for zerolog


hooliganlin:
I'm trying to add span contexts via [apmzerolog](https://github.com/elastic/apm-agent-go/tree/master/module/apmzerolog) and to do that I'd have to add a hook. The weird thing I'm encountering is that if I started a span or some other context to be appended in my logs, I'd have to call the `Hook` method to append the context to the logger for every log event. 

The main question is zerolog suppose to only add hooks on initialization of a `Logger`? 

I created a wrapper over a `zerolog.Logger` and a method called `TraceContext` so that I can do something like:

```
logger.TraceContext(ctx).Info().Msg("successful message")
```
However, I have to attach a hook which creates a new value-based `zerolog.Logger` and then switch the semantics to return a pointer to a `zerolog.Logger` in order to chain it correctly with the other `Events` since the receiver for methods like `Info()` requires a `*zerolog.Logger`.

Example code:
```golang
// Logger wraps a zerolog.Logger as an adapter.
type Logger struct {
	underlying zerolog.Logger
}

func NewLogger(logger zerolog.Logger) *Logger {
	return &Logger{
		underlying: logger,
	}
}

// TraceContext attaches a context.Context to the underlying zerolog.Logger and attaches an apm trace hook.
func (l *Logger) TraceContext(ctx context.Context) *zerolog.Logger {
	// TODO The switching of pointer and value semantic might go against what zerolog was trying to convey,
	// but we need figure out if hooks are meant to be only initialized once vs being added dynamically
	logger := l.underlying.Hook(apmzerolog.TraceContextHook(ctx))
	return &logger
}
```

danielbprice:
I believe that this is probably resolved now that #559 has been merged.  I am using it for Datadog.

danielbprice:
Hi @takanuva15 

Sadly the company for which I wrote this code no longer exists, and so a bunch of the context in brain from this time is just... gone.  Here is what I found in my notes:

```
type ddTraceLoggerHook struct{}

func (d ddTraceLoggerHook) Run(e *zerolog.Event, level zerolog.Level, message string) {
        ctx := e.GetCtx()
        if ctx == nil {
                return
        }
        span, ok := tracer.SpanFromContext(ctx)
        if !ok {
                e.Bool("unified-trace-missing", true)
                return
        }
        e.Str("dd.span_id", strconv.FormatUint(span.Context().SpanID(), 10))
[...]

PR #395, opened 2022-01-07
add context.Context to Event

sams96:
with `WithContext` and `Context` methods to set and get the event context. This addresses issue #290 by allowing you to add a context to an event, which can then be accessed in the hook.

I'm not sure about the naming. `WithContext` and `Context` would match the naming scheme used in `net/http`, but you already have a `Logger.WithContext`, which does the opposite to this (adding a logger to a context) so I'm not sure what would be ideal here.

I'm also not sure what the preferred way to test this would be. I noticed that most of the other event methods don't have their own dedicated tests so I guess for this I could add something to `log_test.go` with the other similar methods, or add something in `hook_test.go` to test the use case.

### DIFF
--- a/event.go
+++ b/event.go
@@ -1,6 +1,7 @@
 package zerolog
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"os"
@@ -27,6 +28,7 @@ type Event struct {
 	stack     bool   // enable error stack trace
 	ch        []Hook // hooks from context
 	skipFrame int    // The number of additional frames to skip when printing the caller.
+	ctx       context.Context
 }
 
 func putEvent(e *Event) {
@@ -770,4 +772,23 @@ func (e *Event) MACAddr(key string, ha net.HardwareAddr) *Event {
 	}
 	e.buf = enc.AppendMACAddr(enc.AppendKey(e.buf, key), ha)
 	return e
+}
+
+// WithContext adds a context to the event
+func (e *Event) WithContext(ctx context.Context) *Event {
+	if e == nil {
+		return e
+	}
+	e.ctx = ctx
+	return e
+}
+
+// Context returns the event context
+//
+// A nil event will return a background context
+func (e *Event) Context() context.Context {
+	if e == nil {
+		return context.Background()
+	}
+	return e.ctx
 }


rs:
The `WithContext` method name gives the assumption that it would return a copy of the `Event` and not a pointer to itself with the `ctx` field changed. I'm not a big fan of using `context` as a bag of data without using it for its main purpose of deadline/cancellation signaling.

sams96:
> The `WithContext` method name gives the assumption that it would return a copy of the `Event` and not a pointer to itself with the `ctx` field changed.

Ah yeah that's a good point. I was looking at the other methods for events which don't create copies. Would you prefer a name change or for it to make a copy?

> I'm not a big fan of using `context` as a bag of data without using it for its main purpose of deadline/cancellation signaling.

It would be extremely convenient for any integration with tracing, since as far as I am aware all of the major tracing libraries for Go use contexts to pass around the trace ids.

danielbprice:
> I'm not a big fan of using context as a bag of data without using it for its main purpose of deadline/cancellation signaling.

Olivier, I see where you are coming from.  But its seems like tracing has become a new and important use case for `context.Context` in the go community.  I don't know how to really avoid it at this point.

I think that the new go-1.21 `slog` interface from the core go team makes it clear that context.Context is endorsed for tracing, and printing trace info is specifically mentioned in their documentation (https://pkg.go.dev/golang.org/x/exp/slog#hdr-Contexts).  The OpenTelemetry process is using `context.Context` for [trace propagation](https://pkg.go.dev/go.opentelemetry.io/otel@v1.16.0/propagation#TraceContext) (as did OpenCensus, Datadog and probably others).

Would you be open to reconsidering this PR?  It would really help.  I think the `InfoCtx()` style of context passing which is present in the Go 1.21 `slog` package also seems like an attractive option, and might in some ways be more ergonomic for developers.

rs:
The `InfoCtx` form would make more sense yes.

danielbprice:
> The `InfoCtx` form would make more sense yes.

I'll see what I can do

sams96:
I'm no longer at the job for which I opened this PR, so I'm not interested in this any more.

PR #406, opened 2022-02-03
MsgFunc function added to Event

### LINKED ISSUE #405: Feature request: Lazy evaluation of the Msg field (see PR #406)
We have trace logging where it is relatively expensive to format the message string.  This is then discarded when the level is lower than trace (ie most of the time)
Whilst this can be mitigated with a level check before logging, it would be much neater if there were a supported function to perform this.
example

``` go
func (e *Event) MsgLazy(createMsg func() string) {
	if e == nil {
		return
	}
	e.msg(createMsg())
}
```

RobRimmer:
Allows lazy evaluation of msg, only if log level is appropriate.
Useful if formatting of log message is expensive and rarely used (e.g. in depth trace logs)
Closes #405

### DIFF
--- a/event.go
+++ b/event.go
@@ -129,6 +129,13 @@ func (e *Event) Msgf(format string, v ...interface{}) {
 	e.msg(fmt.Sprintf(format, v...))
 }
 
+func (e *Event) MsgFunc(createMsg func() string) {
+	if e == nil {
+		return
+	}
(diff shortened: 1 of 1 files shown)


rs:
Please call it `MsgFunc` as all methods of Event are lazy.

PR #407, opened 2022-02-03
journald: don't call Enabled before each write

drakkan:
Enabled opens and close a socket connection by reusing or initializing
a global connection. It does not try to recreate the global socket connection if it failed.
It is useless to call Enabled each time.

I also updated go-systemd to the current development release to include
the following fix:

https://github.com/coreos/go-systemd/commit/75f33b08dbe229fb37b96bf0076907b6b8159af1

This is the only journal related change since the latest stable release.

In SFTPGo I can log to journal conditionally, with the current go-systemd version a socket is always opened even if the journal logging feature is not used, this is the reason I propose to use the latest development version while waiting for a new upstream release

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/rs/zerolog
 go 1.15
 
 require (
-	github.com/coreos/go-systemd/v22 v22.3.2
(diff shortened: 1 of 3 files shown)


drakkan:
Thanks for the really quick merge, much appreciated!

Issue #400, opened 2022-01-24
Context's Logger() method returning a non-pointer


nickcorin:
As per the documentation for `zerolog`, using contexts to pass sub-loggers should work something like this:

```
ctx := log.With().Str("component", "module").Logger().WithContext(ctx)

log.Ctx(ctx).Info().Msg("hello world")

// Output: {"component":"module","level":"info","message":"hello world"}
```

This doesn't work, however, since the `Logger()` method on `Context` returns a non-pointer and `WithContext(context.Context)` requires a pointer receiver.

I have had to use a workaround that looks like this:

```
ctxLogger := log.With().Str("component", "module").Logger()
ctx := &(ctxLogger).WithContext(ctx)

log.Ctx(ctx).Info().Msg("hello world")

// Output: {"component":"module","level":"info","message":"hello world"}
```

It works, but is not pretty.

Were there any major design considerations for using a non-pointer receiver for `Logger()`? 

I'm more than happy to create the PR to change this if you agree that it's a good idea.

smyrman:
This issue is the same as #116, just with a different proposed solution.

mitar:
This has been fixed with #409 and #499 by changing `WithContext` to not be pointer receiver and keeping `Logger` to be value receiver. I think this could be closed unless we want to discuss making all methods be of the same receiver type (also opened in #563).

PR #409, opened 2022-02-04
ctx: Modify WithContext to use a non-pointer receiver

### LINKED ISSUE #116: Chaining WithContext call is broken

nickcorin:
This PR addresses the issues discussed in #116 and #400.

I have modified the receiver for `WithContext` to use a non-pointer receiver and fixed the tests.

The alternative to this, as discussed, would be to modify `Logger()` to return a pointer. If this is the preferred fix, I'd be happy to implement. However, it will have more knock-on effects.

Fixes #116.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -25,9 +25,9 @@ type ctxKey struct{}
 //     l.UpdateContext(func(c Context) Context {
 //         return c.Str("bar", "baz")
 //     })
-func (l *Logger) WithContext(ctx context.Context) context.Context {
+func (l Logger) WithContext(ctx context.Context) context.Context {
 	if lp, ok := ctx.Value(ctxKey{}).(*Logger); ok {
-		if lp == l {
+		if lp == &l {
 			// Do not store same logger.
 			return ctx
 		}
@@ -35,7 +35,7 @@ func (l *Logger) WithContext(ctx context.Context) context.Context {
 		// Do not store disabled logger.
 		return ctx
 	}
-	return context.WithValue(ctx, ctxKey{}, l)
+	return context.WithValue(ctx, ctxKey{}, &l)
 }
 
 // Ctx returns the Logger associated with the ctx. If no logger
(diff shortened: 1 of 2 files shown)


pschwartz:
Watching to see status as I am making use of this and breaking down into multiple statements including a .UpdateContext() is not as clean as the single line.

ctx.go
```diff
@@ -25,17 +25,17 @@ type ctxKey struct{}
 //     l.UpdateContext(func(c Context) Context {
 //         return c.Str("bar", "baz")
 //     })
-func (l *Logger) WithContext(ctx context.Context) context.Context {
+func (l Logger) WithContext(ctx context.Context) context.Context {
 	if lp, ok := ctx.Value(ctxKey{}).(*Logger); ok {
-		if lp == l {
+		if lp == &l {
```
mitar, line 30: So this change makes no sense. Value receiver methods get a new copy of `l` every time they are called. So here you are taking an address of this temporary copied value and comparing it with context pointer. This will always be false. This has been fixed by removing this check in #499.

PR #416, opened 2022-02-21
Support extra arbitrary data at the end of console log

mitar:
I use this to render a stack trace of the error in additional lines.


dobegor:
@rs 
Any objections to this?
It should not impact any existing users, but it will allow for more dev-friendly local use cases such as pretty stacktrace formatting, clickable stacktraces in IDEs from program output are very useful during debugging.

console.go
```diff
@@ -132,6 +134,14 @@ func (w ConsoleWriter) Write(p []byte) (n int, err error) {
 	if err != nil {
 		return n, err
 	}
+
+	if w.FormatExtra != nil {
+		err = w.FormatExtra(evt, buf)
+		if err != nil {
+			return n, err
+		}
+	}
```
rs, line 143: Shouldn't this be called before the \n is added so it can be added to the same line if wanted?

mitar: Sure. My use case was that I add more lines, so this was simpler to do because additional lines (with stack trace) already had newlines.

Maybe the most general approach is to not print `\n` if `w.FormatExtra` is set and is left to `w.FormatExtra` to finalize the line (or lines)?

console.go
```diff
@@ -128,10 +130,18 @@ func (w ConsoleWriter) Write(p []byte) (n int, err error) {
...
-	err = buf.WriteByte('\n')
-	if err != nil {
-		return n, err
+	if w.FormatExtra != nil {
+		err = w.FormatExtra(evt, buf)
+		if err != nil {
+			return n, err
+		}
+	} else {
+		err = buf.WriteByte('\n')
```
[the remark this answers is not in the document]
mitar: I pushed now a version which always prints `\n` at the end.

PR #427, opened 2022-03-19
feat: add hlog CustomFieldHandler

weisdd:
hlog has plenty of useful handlers. What's currently missing is an easy way to enrich log context with a custom field and value, which come from an external source (not from a request itself).

The PR adds a `CustomFieldHandler`. Please, feel free to suggest a better name if needed.


rs:
You can add this field to the logger context instead.

weisdd:
@rs ah, alright, I was thinking only in the context hlog itfself. Should we close the PR then?

PR #435, opened 2022-04-20
add TimestampPrefix

yinyajun:
In order to add time key as a prefix


rs:
I don't understand the purpose. With this, all event will have the same timestamp right?

ikedam:
I suppose what this request want is the same one to what I want.
But this doesn't work as expected. I created #458 .

PR #437, opened 2022-04-26
Adding Event.Type(...) method to log the type of an object

w3aponx:
I found myself wanting to log the type of an object and was convinced there would be a method to do that already, but I didn't find one.  Here's a case where one might want such a method:

```go
	var value interface{}
	var result string
	switch v := value.(type) {
	case int:
		result = strconv.Itoa(v)
	case string:
		result = v
	default:
		log.Panic().Type("type", v).Msg("Unexpected value type")
	}
	log.Info().Msg(result)
```

Let me know if there's anything else that needs to change for this to get accepted.

### DIFF
--- a/internal/json/types.go
+++ b/internal/json/types.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math"
 	"net"
+	"reflect"
 	"strconv"
 )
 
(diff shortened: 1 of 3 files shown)


internal/json/types.go
```diff
@@ -369,6 +369,11 @@ func (e Encoder) AppendInterface(dst []byte, i interface{}) []byte {
 	return append(dst, marshaled...)
 }
 
+// AppendType appends the parameter type (as a string) to the input byte slice.
+func (e Encoder) AppendType(dst []byte, i interface{}) []byte {
+	return e.AppendString(dst, fmt.Sprintf("%T", i))
```
rs, line 374: Please use reflect.TypeOf(i).String() directly instead.

rs (changes requested):

internal/json/types.go
w3aponx: Good suggestion. Had to explicitly handle nil but otherwise pretty straightforward.

w3aponx:
@rs are you waiting for any more changes here?  Just wanted to make sure because AFAIK the requested changes were implemented a while back and I'm not sure how I can convey that any better than by requesting another review.  If it's just that you've been busy with other stuff, no worries -- just wanted to ensure you weren't waiting for something on my part.

PR #441, opened 2022-06-05
Release a new GitHub Releases automatically when a Tag is pushed

### LINKED ISSUE #421: Public release and tagging
When you go on the repository page, you can see that there is only one release : v1.0.0 on 23 June 2017

This can make think a lot of people that zerolog is an old and not very active project, but if we take a look we can really see how active is the project. Almost on tag per month !

It might be a good thing to automatically create release on tag with GitHub workflow

GalvinGao:
This PR fixes #421

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,22 @@
+name: Release on Git Tag
+
+on:
+  push:
+    # trigger on version tag push
+    tags:
+      - "v*"
+
+concurrency:
+  group: ${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: "marvinpinto/action-automatic-releases@v1.2.1"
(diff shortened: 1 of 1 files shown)


yogi1990:
@rs Can you review and merge this PR. This is something we are eagerly looking for it.

rs:
Why do you need github releases?

yogi1990:
@rs the github release definitely gives confidence while evaluating logging/ any package that package is well maintained by community / contributors.  Ideally tags created is enough from technically point of view. But not publishing github release adds some questions on maintainability of package and hiccups in consuming same.

Request is to manually publish github release for now. That will help a lot.

rs:
Well, if there is no actual requirement for gh release, I will stay with simple tags, it’s good enough.

kittydoor:
On top of making releases more visually accessible at first sight (especially when there is existing releases already), they also make it easier to stremline checking software versions across a variety of technologies without having to specially filter for this uses tags, that uses dockerhub, etc etc. They would be very helpful and appreciated.

Issue #316, opened 2021-05-07
Log viewer


GaikwadPratik:
Hi,

First of all, awesome library. 

This is more of question than anything else. Do we have any viewer, like [bunyan](https://github.com/trentm/node-bunyan) to read logs from file in color if they are not being written to console or terminal? If not, how hard would it be to make this compatible with [bunyan](https://github.com/trentm/node-bunyan)?

GaikwadPratik:
@rs ,

Any thoughts or suggestions?

mitar:
Existing pretty-printing console functionality is already taking JSON as input before pretty-printing it. I think it would be easy to make an utility around that.

themarolt:
Just learning go for the past few months and I used bunyan and their CLI piping before so I hope I can make this happen :)

PR #449, opened 2022-07-06
CLI utility to pipe JSON logs through to pretty print and colorize them

themarolt:
Implemented a small utility tool to use when developing an application using zerolog as mentioned in #316.

The only thing that is untested is `Windows` instructions specifically commands that redirect `stderr` to `stdout`:

```shell
some_program_with_zerolog 2>&1 | go run cmd/pretty_pipe/pretty_pipe.go
```

Can someone verify this is working? I only have access to a Linux machine currently.

### DIFF
--- a/cmd/pretty_pipe/README.md
+++ b/cmd/pretty_pipe/README.md
@@ -1,0 +1,81 @@
+# Zerolog Pretty Pipe
+
+This is a basic CLI utility that will colorize and pretty print your JSON logs.
+
+## Problem
+
+When using zerolog it's sometimes difficult to do an environment based log formatting. 
+For production purposes it is useful to have JSON logs to pipe them through Kibana for efficient log entry analysis.
+
+However, when using such applications in development environment there is little use for structured JSON logs.
+Humans are visual beings so the preferred format should be ordinary `Console` format with colors if possible.
+
+Zerolog ships with `ConsoleWriter` which does exactly that. But modifying your code to get pretty printed and colorized
+logs is somewhat difficult.
+
+My solution to do this programmatically was something like this:
+
+```go
+func init() {
+	if !IsProductionEnv() {
+		log.Logger = log.Output(zerolog.NewConsoleWriter())
+	}
+}
+
+func GetLogger(component string) zerolog.Logger {
+	return log.With().
+		Str("Component", module).
+		Logger()
+}
+```
+
+Here `init` is called automatically before `main` by go. `GetLogger` function is used to initialize new loggers.
+If you do something like this:
+```go
+func init() {
+	if !IsProductionEnv() {
+		log.Logger = log.Output(zerolog.NewConsoleWriter())
+	}
+}
+
+var logger = log.With().Str("Component", "Module").Logger()
+```
+
+It won't work since global variables are initialized before `init` is called.
+
+## Solution
+
+A basic CLI tool to pipe application output through it to pretty print & colorize JSON logs to console friendly logs.
+
+## Usage
+
+You can compile it or run it directly. The only issue is that by default Zerolog does not output to `stdout` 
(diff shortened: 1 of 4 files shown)


cmd/pretty_pipe/README.md
```diff
@@ -0,0 +1,81 @@
+# Zerolog Pretty Pipe
```
rs, line 1: What about calling this command prettylog instead?

cmd/pretty_pipe/pretty_pipe.go
```diff
@@ -0,0 +1,35 @@
...
+
+func main() {
+	if !isInputFromPipe() {
+		fmt.Println("The command is intended to work with pipes.")
+		fmt.Println("Usage: app_with_zerolog |  2> >(pretty_pipe)")
+		os.Exit(1)
+		return
+	}
+
+	scanner := bufio.NewScanner(os.Stdin)
```
rs, line 23: You don't need a buffer as zerolog guarantees to perform one write() per event.

rs (changes requested):

cmd/pretty_pipe/README.md
themarolt: Agreed.

cmd/pretty_pipe/pretty_pipe.go
themarolt: I replaced Scanner with Reader as it has a much smaller buffer (4k vs 64k). I hope this is what you meant - otherwise I would need some more explanation :).

rs: Sorry, I meant you don't need a scanner, just a single buffer, each read will get a single event.

themarolt: I think this is ok then. When creating a reader one 4k buffer is allocated and in the loop the content is taken out into a `strings.Builder`.

rs: I'm just saying you don't need a scanner to find the \n boundaries. Just pass the read buffer to the console writer. You can even `io.Copy(consoleWriter, io.Stdin)`.

themarolt: Done

cmd/prettylog/README.md
```diff
@@ -0,0 +1,86 @@
...
+## Problem
+
+When using zerolog it's sometimes difficult to do an environment based log formatting.
+For production purposes it is useful to have JSON logs to pipe them through Kibana for efficient log entry analysis.
+
+However, when using such applications in development environment there is little use for structured JSON logs.
+Humans are visual beings so the preferred format should be ordinary `Console` format with colors if possible.
+
+Zerolog ships with `ConsoleWriter` which does exactly that. But modifying your code to get pretty printed and colorized
+logs is somewhat difficult.
```
rs, line 14: A simpler solution to the problem described here is to detect if the stdout is a terminal and automatically enable console mode if it is. This is a trick I may add as conditional writer in the lib.

I still think this tool is useful when you need to tail zerolog produced output in prod for instance, or in settings when you need both JSON and pretty output.

Could you change the text of the readme to remove the part about why you need this tool so it does not come as the recommended solution to solve this problem?

rs (changes requested):

cmd/prettylog/README.md
themarolt: Yes - of course. I updated the README.md - now it's just a description and usage.

Sorry for taking so long - was on vacation during the summer :)

cmd/pretty_pipe/pretty_pipe.go
mitar: I tested this tool and I think scanner would still be needed because JSONs might be coming from a file - maybe you stored logs to a file and now you are inspecting the file. So `cat log.jsonl | prettylog` and this does not work currently.

Issue #438, opened 2022-05-04
RawJSON doesn't trim newline


byrnedo:
If you're adding a json response that was written using `json.NewEncoder(w).Encode(..)`, there's a trailing newline (by design).
The newline is then left un-escaped in the log entry if you use `RawJSON` to attach it.
Could zerolog trim the content from newlines for us?

rs:
RawJSON does not make any check, the caller is responsible for sending the right input.

byrnedo:
The thing is, isn't that technically a valid json response?

rs:
Not in this context. Basically this method gives you the freedom to add some raw bytes. You are responsible to make the final output valid.

mitar:
Yes, this is the issue of `Encode` adding newlines which is different from `json.Marshal`.

PR #450, opened 2022-07-10
Add New Field: CleanJSON (#438)

VagueCoder:
CleanJSON is just a convenience field that removes newline character `\n` for now.


rs:
I don't feel like this feature is worth increasing the API surface of the library.

Issue #398, opened 2022-01-21
add loglevel for stdout/stderr captured in sub-process sublogger?


pwntr:
**My goal:** I would like to add a loglevel to all outputs from my captured subprocesses stdout/stderr.

Given the following example:

```go
func startSubprocess(pathToBinary string, args []string, subProcessID int) {
    cmd := exec.Command(pathToBinary, args...)

    cmd.Stderr = log.With().Str("subprocessID", strconv.Itoa(subProcessID)).Logger()
    cmd.Stdout = cmd.Stderr
		
    err := cmd.Start()
    // ...
    err = cmd.Wait()
    // ...
}	

func main() {
    // assume global logger existing
    log.Debug().Str("component", "test").Msg("lorem ipsum")
    // startSubprocess(...) a bunch of times
}
```

**As-is:** my output turns out ot be something like this, with no loglevel present for the sub-process outputs:
```json
{"level":"debug","component":"test","time":1642768180137,"message":"lorem ipsum"}
{"subprocessID":"1","time":1642768180138,"message":"any output from my sub-process received on stdout or stderr"}
```

**To-be:** add a (fixed/given) loglevel like `"level":"info"` to be added to every line received from the stdout/stderr from the sub-processes:

```json
{"level":"debug","component":"test","time":1642768180137,"message":"lorem ipsum"}
{"level":"info","subprocessID":"1","time":1642768180138,"message":"any output from my sub-process received on stdout or stderr"}
```

I'm revisiting this after a few months and currently try to wrap my head around the `io.Writer`, which _might_ be of help here?

Any ideas or easier solutions and pointers are greatly appraciated, thanks a lot for your help :)!

pwntr:
For a pure json ouput I _kinda_ worked around my problem in a hacky way by just adding the loglevel via `.Str()` and checking for it manually like so:

```go
if zerolog.GlobalLevel() <= zerolog.InfoLevel {
    cmd.Stderr = log.With().Str("level", zerolog.InfoLevel.String()).Str("subprocessID", strconv.Itoa(subProcessID)).Logger()
    cmd.Stdout = cmd.Stderr
}
```

However, this unfortunately breaks the pretty `ConsoleWriter` output, if one so desires to use it ocassionally:

```
...
6:43PM TRC some parent process output on trace level
6:43PM INF first line of sub-process / child output captured
second line of sub-process / child output captured
another line of sub-process / child output captured
another line of sub-process / child output captured
another line of sub-process / child output captured
[...]

PR #455, opened 2022-07-24
Field LogLevel at Logger, Context & Event Levels

VagueCoder:
Field `LogLevel` explicitly adds/replaces the log level to logs as per previous `Logger`, `Context` or `Event`. This only helps in printing and not changing log level.

1. If the item with key `level` doesn't exist, `LogLevel` will add new with log level string as value.
2. If `level` already exists but value differs, this will replace value.
3. If both key and value match, no change.

This will help in closing #398


rs:
I'm going to refuse this PR:
* parsing the buffer is expensive (especially with regex)
* log level is not respected, making this feature confusing
* it is to much complexity added (I will have to maintain that) for the benefit

PR #457, opened 2022-07-29
pass program counter to CallerMarshalFunc

### LINKED ISSUE #456: Pass PC to caller marshaller to allow more detailed information
Here the program counter var is `_` ignored:
https://github.com/rs/zerolog/blob/4099072c03f2f4e61fa08f70adf9a25983f0cd8e/event.go#L743-L753

If it was passed into the marshaller, we could also print the method name, which we're currently missing.

As this would be a backwards-incompatible change, I'm creating this issue first before creating a PR, but I think it would be worth it and very easy to fix (for the few people that actually overwrite that method manually)

Edit:
This is how you could get the func name:
```go
.Str("func", runtime.FuncForPC(pc).Name())
```

Hades32:
This passes the pc var to marshallers to be able to add function names to caller fields.

This is a breaking change for those that have overridden `CallerMarshalFunc`  before, as the func signature changes.

All tests passed locally.

Resolves #456

### DIFF
--- a/log_test.go
+++ b/log_test.go
@@ -799,7 +799,7 @@ func TestCallerMarshalFunc(t *testing.T) {
 			return strings.Join(parts[len(parts)-2:], "/") + ":" + strconv.Itoa(line)
 		}
 
-		return file + ":" + strconv.Itoa(line) + " " + runtime.FuncForPC(pc).Name()
+		return runtime.FuncForPC(pc).Name() + ":" + file + ":" + strconv.Itoa(line)
 	}
 	pc, file, line, _ = runtime.Caller(0)
 	caller = CallerMarshalFunc(pc, file, line+2)
--- a/log_test.go
+++ b/log_test.go
@@ -782,7 +782,7 @@ func TestCallerMarshalFunc(t *testing.T) {
 
 	// test default behaviour this is really brittle due to the line numbers
 	// actually mattering for validation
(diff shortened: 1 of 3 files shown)


log_test.go
```diff
@@ -793,16 +793,16 @@ func TestCallerMarshalFunc(t *testing.T) {
...
 	defer func() { CallerMarshalFunc = origCallerMarshalFunc }()
-	CallerMarshalFunc = func(file string, line int) string {
+	CallerMarshalFunc = func(pc uintptr, file string, line int) string {
 		parts := strings.Split(file, "/")
 		if len(parts) > 1 {
 			return strings.Join(parts[len(parts)-2:], "/") + ":" + strconv.Itoa(line)
 		}
 
-		return file + ":" + strconv.Itoa(line)
+		return file + ":" + strconv.Itoa(line) + " " + runtime.FuncForPC(pc).Name()
```
rs, line 802: Shouldn't it be func:file:line?

Hades32: Yes, that's nicer. I didn't put much thought into it, as it's only used locally within the test TBH. In my personal use-case I probably won't even include the file name.

Hades32:
thanks for merging @rs! Would you mind also creating a new git tag, so I can update our deps?

Hades32:
@rs could you create a v1.28.0 tag?

rs:
done

Issue #462, opened 2022-08-16
Not getting the stack in an error


conor-tread:
Im using the package in an AWS serverless codebase and when I force an error Im not seeing the stack trace in the logs. In the init() function I have zerolog.ErrorStackMarshaler = pkgerrors.MarshalStack. 

I can see my error message in the log that gets printed its just not printing the stack

Mehrdad-Dadkhah:
+1

skormos:
You need to use [WithStack()](https://pkg.go.dev/github.com/pkg/errors#WithStack) from [pkg/errors](https://github.com/pkg/errors). 

The documentation lists that as a requirement: https://github.com/rs/zerolog#error-logging-with-stacktrace

skormos:
I created another issue requesting `pkg/errors` be [removed as a dependency](https://github.com/rs/zerolog/issues/470) since that library is now in maintenance mode.

mrwormhole:
+1 same issue, it used to work in September 2022 but no longer works, I wonder if this is something related to dependencies of zerolog, zerolog v1.27.0 was working for me if ErrorStackMarshaler is specified or not. Now both specified or not ways are working

@skormos your solution only prints line number log is called, not where the err returned from so it is basically not stack but just line number printer for log

mrwormhole:
oh well ignore me, it works, It only works on pkg/errors not on normal errors for Stack() calls

xsteadfastx:
> You need to use [WithStack()](https://pkg.go.dev/github.com/pkg/errors#WithStack) from [pkg/errors](https://github.com/pkg/errors).
> 
> The documentation lists `pkg/errors` as a requirement: https://github.com/rs/zerolog#error-logging-with-stacktrace
> 
> Not sure if there are any downsides to the below approach, but you can just wrap the `MarshalStack` reference in a closure, so you don't need to pollute the entire codebase with `pkg/errors` imports, and only use this reference when you add the `.Stack()` call to the logger context chain:
> 
> ```go
> import (
> 	"github.com/pkg/errors"
> 	"github.com/rs/zerolog"
> 	"github.com/rs/zerolog/pkgerrors"
> )
> ...
> 	zerolog.ErrorStackMarshaler = func(err error) interface{} {
> 		return pkgerrors.MarshalStack(errors.WithStack(err))
> 	}
> ```

could you explain why we need this? your code just works fine, but why its not in the official code? i couldnt get it to work with the example out of the readme. should we create a PR to add your fix?

[...]

PR #471, opened 2022-09-07
(#470) Localize github.com/pkg/errors

### LINKED ISSUE #470: Stack: remove `github.com/pkg/errors` as a dependency
The `.Stack()` context method depends on [pkg/errors](https://github.com/pkg/errors) which is now in Maintenance mode. 

Recommend removing this dependency, and bringing the stack functionality directly into this project.

vaguecoder:
# Motivation
[Issue](https://github.com/rs/zerolog/issues/470) #470 reported by @skormos

# Changes
Adds new internal package `errors` with test coverage.

# Summary
Localizes following functions (and required types and methods) from [github.com/pkg/errors](https://github.com/pkg/errors) as it is no longer being maintained.
```go
New(message string) error
Wrap(err error, message string) error
```
# Related Issues
Closes #470
Helps in #462

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,7 +411,7 @@ log.Info().Msg("hello world")
 Equivalent of `Lshortfile`:
 
 ```go
-zerolog.CallerMarshalFunc = func(file string, line int) string {
+zerolog.CallerMarshalFunc = func(pc uintptr, file string, line int) string {
     short := file
     for i := len(file) - 1; i > 0; i-- {
         if file[i] == '/' {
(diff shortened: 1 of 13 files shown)


rs:
After this change, as we do not expose the errors.Wrap, how are users supposed to use this feature?

vaguecoder:
@rs I totally missed that `Wrap` is required by users. Can it be included in the same package as MarshalStack? I understand that it means extending `pkgerrors` layer.

rs:
Reading the change I'm not really fan of the resulting API. I wonder if we could keep backward compat (when getting a wrapped error) and at the same time, auto-wrap an error when Stack is present and Err is called. This way we do not expose a new API.

PR #477, opened 2022-09-12
Use x/tools/go/packages to load package info.

godfried:
The loader package currently used to retrieve package info is deprecated. This change switches to the recommended x/tools/go/packages.


mitar:
The linter from this repository has been recently deprecated in favor of https://github.com/ykadowak/zerologlint. I think this PR should thus not be merged.

PR #490, opened 2022-10-08
feat: added support for go-errors stack traces

alesbrelih:
Added support to print out stack traces generated by the [go-errors](https://github.com/go-errors/errors) library.

I've put the new code in goerrors package, similar to the existing pkgerrors package, becase I wasn't sure what kind of folder structure would you prefer. If you want I can change this structure to something like:
```
stacktraces/{pkgerrors,goerrors}
```

Thank you for this library!

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,7 +232,9 @@ func main() {
 
 #### Error Logging with Stacktrace
 
-Using `github.com/pkg/errors`, you can add a formatted stacktrace to your errors. 
+Zerolog supports two external error libraries to print out stack traces.
+
+Using `github.com/pkg/errors`, you can add a formatted stacktrace to your errors like this:
 
 ```go
 package main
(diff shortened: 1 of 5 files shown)


mitar:
I think zerolog should go away from using external errors libraries to format stack traces but just support extracting stack traces from them. go-errors does implement the `Callers() []uintptr` interface and only that (without having to import go-errors) should be used to extract callers and then process those callers inside zerolog itself, in the same way for all errors.

The [errors package I made](https://gitlab.com/tozd/go/errors) as a v2 of pkgerrors implements `StackTrace() []uintptr` interface. So, zerolog would only have to check few of those interfaces and extract raw callers and then format them themselves (ideally into JSON first, and then from JSON into console).

pkgerrors sadly does not support that, so it has to be imported to obtain the stack trace.

BTW, I would suggest that the order of checks on the error is that first it s checked if error knows how to JSON marshall itself, or even if it supports zerolog's JSON conversion. And if not, then check if they support extracting `[]uintptr` out.

th0th:
Looking forward for this to get merged 🙌

th0th:
Hey @rs, does your closing this PR mean the feature is rejected? 🙃

rs:
Yes, I will not be adding any dependencies to this library.

Issue #498, opened 2022-10-20
Invalid condition in WithContext method.


tep:
Since non-pointer receivers are passed **_by value_**, once the receiver for `WithContext` was reverted to a non-pointer, [this condition](https://github.com/rs/zerolog/blob/89617ff99b2feb29ab2d8b6258fac9d671073792/ctx.go#L30) is guaranteed to always be false.

In other words, the receiver (`l`) is a _copy_ of the Logger struct from the caller and its address can never be the same as the logger attached to the `Context`. 

Either the receiver should be converted back to a pointer (and different fixes devised for #116 and #400) -- or, this condition should simply be removed and the docs updated to say that the Context will always be replaced no matter what (because that's the actual behavior with the current logic).

[Here's a [Go Playground link](https://go.dev/play/p/Abw-heqUhwP) proving that non-pointer receivers have different addresses from those of their caller]

rs:
This is a regression from this commit: https://github.com/rs/zerolog/commit/588a61c2df4bf7b92e12b826b226059b9f262190

rs:
Perhaps we should remove the condition altogether.

tep:
That's what I'd originally done in PR #499 but then changed it to add back a similar condition using `reflect.DeepEqual` but, I can always pull it out again.

rs:
Ok, I'll prefer removing the condition then.

tep:
Done. 

#499 should be ready to go.

PR #499, opened 2022-10-21
Fix docs and behavior of WithContext

tep:
This change updates `WithContext` to better match documented behavior and also updates the documentation to both clarify existing behavior as well as the behavior expected by existing tests.

Note: since `WithContext` no longer employs a pointer receiver, the value comparison against the Context's attached logger value must be done via `reflect.DeepEqual`.

Fixes: #498

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -14,10 +14,15 @@ func init() {
 
 type ctxKey struct{}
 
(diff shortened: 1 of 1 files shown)


jpalawaga:
Thank you for this. Was bitten by the docs not quite matching the behaviour in 1.2.6.

PR #515, opened 2023-01-07
Add Any field function

### LINKED ISSUE #514: Add Any field function for any value
Go introduced `any` type, that is just  an empty interface under the hood. I think it would be good to introduce `Any()` function in zerolog as well. Under the hood it will use the old `Interface()` function.

Example:
```go
// before
log.Debug().Interface("params", p).Msg("charge request")

// after
log.Debug().Any("params", p).Msg("charge request")
```

It makes the line a bit shorter and, in my opinion, makes it more obvious of what it's going on log.

Also I don't propose to remove`Interface()` function, I propose to add a new function that's going to do the same thing that the old function did.

dils2k:
resolves #514

### DIFF
--- a/event.go
+++ b/event.go
@@ -707,6 +707,11 @@ func (e *Event) TimeDiff(key string, t time.Time, start time.Time) *Event {
 	return e
 }
 
+// Any is a wrapper around Event.Interface.
+func (e *Event) Any(key string, i interface{}) *Event {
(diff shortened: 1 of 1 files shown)


dils2k:
@rs

electrofocus:
Not a bad idea, it looks nice

PR #523, opened 2023-02-13
fix level parser

mvrahden:
This PR fixes level parsing for any string casing

Docs-formatting was a result of Go-fmt.

### DIFF
--- a/log.go
+++ b/log.go
@@ -161,24 +161,24 @@ func (l Level) String() string {
 // ParseLevel converts a level string into a zerolog Level value.
 // returns an error if the input string does not match known values.
 func ParseLevel(levelStr string) (Level, error) {
-	switch strings.ToLower(levelStr) {
-	case LevelFieldMarshalFunc(TraceLevel):
+	switch {
+	case strings.EqualFold(levelStr, LevelFieldMarshalFunc(TraceLevel)):
 		return TraceLevel, nil
-	case LevelFieldMarshalFunc(DebugLevel):
+	case strings.EqualFold(levelStr, LevelFieldMarshalFunc(DebugLevel)):
 		return DebugLevel, nil
-	case LevelFieldMarshalFunc(InfoLevel):
+	case strings.EqualFold(levelStr, LevelFieldMarshalFunc(InfoLevel)):
 		return InfoLevel, nil
-	case LevelFieldMarshalFunc(WarnLevel):
+	case strings.EqualFold(levelStr, LevelFieldMarshalFunc(WarnLevel)):
 		return WarnLevel, nil
-	case LevelFieldMarshalFunc(ErrorLevel):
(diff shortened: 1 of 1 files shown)


log.go
```diff
@@ -162,23 +161,23 @@ func (l Level) String() string {
 // returns an error if the input string does not match known values.
 func ParseLevel(levelStr string) (Level, error) {
 	switch strings.ToLower(levelStr) {
-	case LevelFieldMarshalFunc(TraceLevel):
+	case strings.ToLower(LevelFieldMarshalFunc(TraceLevel)):
```
rs, line 164: what about using `EqualFold` instead of double `ToLower`?

mvrahden: As far as I'm aware `EqualFold` performs a check (returns a `bool`, so it wouldn't be a drop in replacement for `ToLower`, but we would rewrite the logic with an individual `if`-statement for each case.

rs: That’s fine

mvrahden:
Is there any chance, that we can merge this and release?

log.go
```diff
@@ -161,24 +160,31 @@ func (l Level) String() string {
 // ParseLevel converts a level string into a zerolog Level value.
 // returns an error if the input string does not match known values.
 func ParseLevel(levelStr string) (Level, error) {
-	switch strings.ToLower(levelStr) {
-	case LevelFieldMarshalFunc(TraceLevel):
+	if strings.EqualFold(levelStr, LevelFieldMarshalFunc(TraceLevel)) {
 		return TraceLevel, nil
-	case LevelFieldMarshalFunc(DebugLevel):
+	}
+	if strings.EqualFold(levelStr, LevelFieldMarshalFunc(DebugLevel)) {
```
rs, line 166: please keey using a switch/case with the condition.

mvrahden: You mean I should revert the last commit?
Because as I explained before the `EqualFold` does not work with `switch` statement.

rs: Use the EqualFold as the expression of the switch with an empty switch.

mvrahden: Alright, the expressions were exchanged.

log.go
```diff
@@ -2,97 +2,96 @@
...
 //
-//     // And use it:
-//     var h SeverityHook
-//     log := zerolog.New(os.Stdout).Hook(h)
-//     log.Warn().Msg("")
-//     // Output: {"level":"warn","severity":"warn"}
+//	// And use it:
+//	var h SeverityHook
+//	log := zerolog.New(os.Stdout).Hook(h)
+//	log.Warn().Msg("")
```
rs, line 82: Please remove all the space changes above

mvrahden:
The change should now be ready to merge, right?

mvrahden:
@rs any updates from your side?

PR #527, opened 2023-02-27
feat: add `GetKeys` to get event keys

julienrbrt:
As far as I know, there is no way to get all the keys of a particular event.
This is handy (even though expensive), to use in hooks.

### DIFF
--- a/event.go
+++ b/event.go
@@ -1,6 +1,7 @@
 package zerolog
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 	"os"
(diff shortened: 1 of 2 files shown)


julienrbrt:
Gently pinging you @rs, this would greatly help us.

rs:
This won’t work until event is “closed” (sent) and won’t work with bson encoding. I’m not for adding/maintaining this. Thanks for the suggestion.

julienrbrt:
> This won’t work until event is “closed” (sent) and won’t work with bson encoding. I’m not for adding/maintaining this. Thanks for the suggestion.

This makes sense, thank you review! Would you know a solution for reaching the same goal?

PR #534, opened 2023-03-28
doc: update readme to use os-specific path separator

youming-lin:


andrii-zakurenyi:
Hi @youming-lin 
Can you please update the code to use `filepath.Base()` instead of that custom for loop: https://pkg.go.dev/path/filepath#Base

Something like:
```go
return filepath.Base(file) + ":" + strconv.Itoa(line)
```

youming-lin:
@andrii-zakurenyi 
Absolutely! That's what I ended up using as well after a bit more Googling 😃 

I've updated the PR, please check again.

PR #535, opened 2023-03-31
Add Event.Size() method to return the size of the event buffer

bderrly:
I am writing a hook to increment some opentelemetry metrics in a logging library. One of the things I would like to count is the sum of bytes written. I can easily get the count of bytes for the "message" (as a parameter to the hook callback) but I want the entire event that will be written.

I wasn't sure if it would be preferable to return a zero or -1 in the case of a nil Event. Thoughts?

### DIFF
--- a/event.go
+++ b/event.go
@@ -83,7 +83,7 @@ func (e *Event) write() (err error) {
 }
 
 // Size returns the size of the buffer in the Event.
-// This is most likely useful inside a Hook.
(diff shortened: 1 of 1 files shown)


rs:
Some other hooks after yours could add bytes and the closing of the event will add some too. I would suggest to use a io.Writer wrapper on the output to log event size. Each call to write is guaranteed to be a single full event.

PR #541, opened 2023-04-14
FormatPartValueByName for flexible custom formatting for ConsoleWriter

sted:


PeterBocan:
This is very nice addition to zerolog. Any concerns against having this feature?

I have the same issue, where I would like to have a contextual value to be right after the timestamp. Right now it's not possible to achieve this.

cc. @rs

PR #550, opened 2023-05-06
Add fields order

madkins23:
After I did this work I saw PR 170 which is apparently languishing. My bad for not checking first. On the other hand this PR currently has no merge issues and I have time to respond to comments.

### DIFF
--- a/console.go
+++ b/console.go
@@ -66,7 +66,7 @@ type ConsoleWriter struct {
 	// FieldsOrder defines the order of contextual fields in output.
 	FieldsOrder []string
 
-	fieldIsOrdered map[string]bool
+	fieldIsOrdered map[string]int
 
 	// FieldsExclude defines contextual fields to not display in output.
 	FieldsExclude []string
@@ -190,10 +190,11 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
 		}
 		fields = append(fields, field)
 	}
-	sort.Strings(fields)
 
 	if len(w.FieldsOrder) > 0 {
-		fields = w.orderFields(fields)
+		w.orderFields(fields)
+	} else {
+		sort.Strings(fields)
 	}
 
 	// Write space only if something has already been written to the buffer, and if there are fields.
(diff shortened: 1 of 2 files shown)


console.go
```diff
@@ -165,20 +175,34 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
...
 			continue
 		}
+
+		var isOrdered bool
+		for _, ordered := range w.FieldsOrder {
+			if field == ordered {
+				isOrdered = true
+				break
+			}
+		}
```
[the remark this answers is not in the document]
madkins23: I agree and would have done it that way (and maybe saved the map on the `ConsoleWriter` for reuse) but the surrounding code was just iterating in a loop (see `FieldsExclude` just above) and I didn't want to break with the original author's style. Or worse yet start randomly fixing the existing loop(s).

That said, I don't mind making any reasonable change that will help the PR get accepted and merged.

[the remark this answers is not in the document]
madkins23: Got a round toit.

.gitignore
FlandiaYingman, line 1: I think you should remove this commit, because this is irrelevant with the feature we want to implement.

go.mod
FlandiaYingman, line 1: The dependencies should not be updated by an feature implementing pull request for same reason: this is irrelevant with the feature we want to implement.

FlandiaYingman (changes requested):

.gitignore
madkins23: Done

go.mod
madkins23: Done

rs:
Could you please make this change not allocating additional memory on each write if the feature is not used?

madkins23:
Good idea. That also cleaned up the code a lot. Not sure what I was thinking when I did all that before. :-(

I stored a `map[string]bool` on the `ConsoleWriter` to speed up subsequent calls. It won't get created until the first call to `orderFields` so it's just a wasted pointer otherwise, and `ConsoleWriter` shouldn't be created a bunch of times unless I'm missing something.

BTW, there's another similar situation with `FieldsExclude` at the beginning of `ConsoleWriter.writeFields` that could be solved the same way.  The current solution iterates over `w.FieldsExclude` for each field which feels like `O(n^2)`. I don't mind writing the PR if you agree @rs.

Both of those last two notes assume that `ConsoleWriter` fields don't get changed over time. If that is not true then I need to remove that tweak on this PR. The `w.FieldsExclude` issue might still benefit from being converted to a `map[string]bool` at the beginning of  `ConsoleWriter.writeFields`.

console.go
```diff
@@ -187,6 +192,10 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
 	}
 	sort.Strings(fields)
 
+	if len(w.FieldsOrder) > 0 {
+		fields = w.orderFields(fields)
+	}
+
```
rs, line 198:
```suggestion

	if len(w.FieldsOrder) > 0 {
		fields = w.orderFields(fields)
	} else {
		sort.Strings(fields)
	}

```

console.go
```diff
@@ -318,6 +327,34 @@ func (w ConsoleWriter) writePart(buf *bytes.Buffer, evt map[string]interface{},
...
+			otherFields = append(otherFields, fieldName)
+		}
+	}
+	result := make([]string, 0, len(fields))
+	for _, fieldName := range w.FieldsOrder {
+		if hasOrderedField[fieldName] {
+			result = append(result, fieldName)
+		}
+	}
+	return append(result, otherFields...)
```
rs, line 355:
```suggestion
	if w.fieldIsOrdered == nil {
		w.fieldIsOrdered = make(map[string]int)
		for i, fieldName := range w.FieldsOrder {
			w.fieldIsOrdered[fieldName] = i
		}
	}
	sort.Slice(fields, func(i, j int) bool {
		ii, iOrdered := w.fieldIsOrdered[fields[i]]
		jj, jOrdered := w.fieldIsOrdered[fields[j]]
		if iOrdered && jOrdered {
			return ii < jj
		}
		if iOrdered {
			return true
		}
		if jOrdered {
			return false
		}
		return fields[i] < fields[j]
	})
```

rs (changes requested):
Please consider the suggested changes to avoid allocations.

console.go
madkins23: Done

console.go
madkins23: Nice!

rs (approved):

Issue #558, opened 2023-06-21
Would like to pass the current context to the Event to facilitate Tracing


danielbprice:
I'm trying to get zerolog hooked up to Datadog's unified tracing/logging scheme.  In this scheme, you tag your log messages with the current trace & span IDs, and then in the Datadog console you can navigate between them.  When it works, it's pretty awesome.

If anyone has a recipe for this, I'd love to know about it.  I am not sure, but I think there is a facility missing in zerolog to make this work smoothly.

Datadog needs you to publish special fields in your log messages: `dd.trace_id`, `dd.span_id`, `dd.service`, `dd.env` and `dd.version`.  Some of these (`service`, `env`, and `version`) are straightforward and static during the runtime of the application.  But `dd.trace_id` and `dd.span_id` come from your current span.

Generally speaking, at least for Datadog, the span rides around in your `context.Context`, and functions that want to do something to the span, or make a new child span, will reference the context to get it.  You could certainly extract that from your current context, and manually add those to each and every zerolog log message in your application, i.e. with `.Str("dd.trace_id", trace_id).Str("dd.span_id", span_id)` but that would really suck.

zerolog provides a `Hook` function, and a hook function seems like a pretty good place to add in the trace ID and the span ID.  The hook could be defined at the program's startup and added to the top-level logger.  But this won't work because there is no obvious way to get the current `context.Context` into the Hook, that I can find.

It appears that in go 1.21, when the Go team defined the `slog` API, they solved this problem by creating `Ctx` variants of their top-level functions, because "_Some handlers may wish to include information from the [context.Context](https://pkg.go.dev/context#Context) that is available at the call site. **One example of such information is the identifier for the current span when tracing is enabled**._" See https://pkg.go.dev/golang.org/x/exp/slog#hdr-Contexts.

And so essentially my request is for a way to do this in zerolog, because I don't want to port all of my logging to slog.  My hope is to be able to do something like:

```
logger := zerolog.New(os.Stderr).Hook(tracingHook)
...
logger.InfoCtx(ctx).Msg("hello world")
```

Or perhaps: `logger.Info().Ctx(ctx).Msg("hello world")`

If there is already a way to do this in zerolog, I've had a lot of trouble finding it; in that case, this is a request for a "howto" document somewhere.  I read all about Logger.WithContext and Logger.Ctx but as best I can tell they don't do what I'm asking for here.  Another options would be something like `Event.Ctx(ctx context.Context)`.

[...]

PR #559, opened 2023-06-22
Allow callers to pass go context through to hooks

danielbprice:
Add Logger.{Trace,Debug,Info,Warn,...}Ctx() and similar functions to allow go context to propagate to Hooks.  Add Event.Context() to make the context retrievable by hooks.  Facilitates writing hooks which fetch tracing context from the go context.

---
This PR helps to address #395, #558 and maybe #290.  It is modeled off of the similar interfaces in the new `slog` stdlib package in go 1.21.

### DIFF
--- a/event.go
+++ b/event.go
@@ -1,6 +1,7 @@
 package zerolog
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"os"
@@ -24,9 +25,10 @@ type Event struct {
 	w         LevelWriter
 	level     Level
 	done      func(msg string)
-	stack     bool   // enable error stack trace
-	ch        []Hook // hooks from context
-	skipFrame int    // The number of additional frames to skip when printing the caller.
+	stack     bool            // enable error stack trace
+	ch        []Hook          // hooks from context
+	skipFrame int             // The number of additional frames to skip when printing the caller.
+	ctx       context.Context // Optional Go context for event
 }
 
 func putEvent(e *Event) {
@@ -415,6 +417,28 @@ func (e *Event) Stack() *Event {
 		e.stack = true
 	}
 	return e
+}
+
+// Ctx adds the Go Context to the *Event context.  The context is not rendered
+// in the output message, but is available to hooks and to Func() calls via the
+// GetCtx() accessor. A typical use case is to extract tracing information from
+// the Go Ctx.
+func (e *Event) Ctx(ctx context.Context) *Event {
+	if e != nil {
+		e.ctx = ctx
+	}
+	return e
+}
+
+// GetCtx retrieves the Go context.Context which is optionally stored in the
+// Event.  This allows Hooks and functions passed to Func() to retrieve values
+// which are stored in the context.Context.  This can be useful in tracing,
+// where span information is commonly propagated in the context.Context.
+func (e *Event) GetCtx() context.Context {
+	if e == nil || e.ctx == nil {
+		return context.Background()
+	}
+	return e.ctx
 }
 
 // Bool adds the field key with val as a bool to the *Event context.
--- a/README.md
+++ b/README.md
@@ -513,25 +513,53 @@ stdlog.Print("hello world")
 
 ### context.Context integration
 
-The `Logger` instance could be attached to `context.Context` values with `logger.WithContext(ctx)`
-and extracted from it using `zerolog.Ctx(ctx)`.
+Go contexts are commonly passed throughout Go code, and this can help you pass
+your Logger into places it might otherwise be hard to inject.  The `Logger`
+instance may be attached to Go context (`context.Context`) using
+`Logger.WithContext(ctx)` and extracted from it using `zerolog.Ctx(ctx)`.
+For example:
 
-Example to add logger to context:
 ```go
-// this code attach logger instance to context fields
-ctx := context.Background()
-logger := zerolog.New(os.Stdout)
-ctx = logger.WithContext(ctx)
-someFunc(ctx)
+func f() {
+    logger := zerolog.New(os.Stdout)
+    ctx := context.Background()
+
+    // Attach the Logger to the context.Context
+    ctx = logger.WithContext(ctx)
+    someFunc(ctx)
+}
+
+func someFunc(ctx context.Context) {
+    // Get Logger from the go Context. if it's nil, then
+    // `zerolog.DefaultContextLogger` is returned, if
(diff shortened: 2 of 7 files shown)


event.go
```diff
@@ -185,6 +188,18 @@ func Dict() *Event {
 	return newEvent(nil, 0)
 }
 
+// GetContext allows retrieval of the context.Context which is optionally
+// stored in the event.  This allows Hooks to retrieve values which are stored
+// in the context.Context.  This can be useful in tracing, where span
+// information is commonly propagated in the context.Context.
+// See also Logger.InfoCtx, etc.
+func (e *Event) GetContext() context.Context {
+	if e.context == nil {
```
rs, line 197: Make sure to test for e == nil

rs (changes requested):

event.go
danielbprice: I would assume it would be impossible to get into that situation inside of a hook?  But sure, it seems harmless to check it.

danielbprice:
I added a bunch of testing today, as I didn't want to be less than thorough; however, if it seems like overkill I am happy to yank it.

event.go
```diff
@@ -185,6 +188,18 @@ func Dict() *Event {
 	return newEvent(nil, 0)
 }
 
+// GetContext allows retrieval of the context.Context which is optionally
+// stored in the event.  This allows Hooks to retrieve values which are stored
+// in the context.Context.  This can be useful in tracing, where span
+// information is commonly propagated in the context.Context.
+// See also Logger.InfoCtx, etc.
+func (e *Event) GetContext() context.Context {
```
rs, line 196: `Context` instead of `GetContext` would be more idiomatic.

danielbprice: Good point!

event.go
danielbprice: Fixed

rs:
Thinking about the API, what do you think about, instead of exposing a new `Context` method on the event, add an alternative Hook interface that would take a `ctx` as the first argument?

rs (approved):

danielbprice:
> Thinking about the API, what do you think about, instead of exposing a new `Context` method on the event, add an alternative Hook interface that would take a `ctx` as the first argument?

Sure thing, should be an easy change.  I want you to be satisfied with the interface choices.

danielbprice:
> > Thinking about the API, what do you think about, instead of exposing a new `Context` method on the event, add an alternative Hook interface that would take a `ctx` as the first argument?
> 
> Sure thing, should be an easy change. I want you to be satisfied with the interface choices.

Hi Olivier,

So this turned out to be quite a bit more awkward than I had expected, as it required duplicating almost all of the Hook into a new `HookCtx` interface-- we can't add e.g. `RunCtx` to the Hook interface.  I've put the changes up on a branch for you to see, but it seems awkward to duplicate so much code and mechanism to get this.  https://github.com/danielbprice/zerolog/commit/42e0eb265132d8575e8927babdb2856011664544.   What do you think?

Also, over the weekend, I began to wonder I spent too little time thinking through the interface here.  Would we be better off passing the context directly into the terminal routines-- as in: `MsgCtx(ctx, "hello")` and `MsgfCtx(ctx, "hello %s", "there")`.  Or into a method on the event (ex: `logger.Warn().Str("foo", "bar").Ctx(ctx).Msg("hello")`?   You've had much more time to think about what is right for zerolog, so I am open to refining this some more.  In my mind, there are competing forces:  `.InfoCtx()` feels somewhat more Go-Idiomatic and aligns with `slog`.  `.Ctx()` feels more idiomatic to the method-chaining style and more "native" to zerolog (IMO).

So, three options I see:
1.  (keep what we have in this PR): `logger.InfoCtx(ctx).Str("mystr", "abcd").Msg("hello")`
2. (pass at terminal end): `logger.Info().Str("mystr", "abcd").MsgCtx(ctx, "hello")`
3. (pass via new function): `logger.Info().Str("mystr", "abcd").Ctx(ctx).Msg("hello")`

Any thoughts?  It seems to me that #1 and #3 are the most likely to be the right answer.

danielbprice:
> 
> Any thoughts? It seems to me that either #1 or #3 is most likely to be the right answer.

@rs Just wondered if you have any new thoughts here?  I've just wrapped up another project and might have a day to devote to this in the coming week.

rs:
3. probably makes the most sense. I would probably use `WithContext(ctx)` and `Context()` as method names.

danielbprice:
Hi!  Sorry for the delay, but work has been busy and I've had to deal with Python structured logging. 🤮 😡 🤬 

I posted a significantly revised review:
- After a bunch of experimentation, I propose settling on the use of `Ctx` when referring to the go context.  This is to try to avoid confusion with zerolog.Context.  Also, I foresee typing `log.Info().Ctx(ctx).Msg("whatever")` a _lot_, so the brevity is nice.  Since `ctx` is almost universally the name used for variables of type `context.Context`, I think this will feel natural to most programmers.  I know you suggested using `WithContext(ctx)`.  But it seemed confusing to have the already-existing `Logger.WithContext()` (which is about storing the logger to the go context) and a newly added Event.WithContext() which does something substantially different.
- I have pulled this version into my project at work, deployed it to our development environment, and confirmed that I can implement Datadog unified logging/tracing works with it.
- I also tried to improve the documentation for getting and storing `Logger`s to the Go context, so that new users will have a clear idea of what the two different features are, and when to use them.
- I did not bring in the revised HookCtx interface that I prototyped in https://github.com/danielbprice/zerolog/commit/42e0eb265132d8575e8927babdb2856011664544.  It just seemed really awkward to me, and a lot of code for a relatively small benefit.  If there was ever a `v2` of the interface, then I would suggest revising Hook to take a context.Context.

danielbprice:
Hi @rs,  any chance you could take a look at this (significantly) revised version?  I have some cycles this week for another round of review and fixing, and hopefully integration.

danielbprice:
@rs Thank you so much!  I really appreciate the opportunity to contribute.

inputvalidation:
Really appreciate this PR. I was writing my code against it and then found it was not released yet 😇 Not sure of zerolog's release policy but would be great to be able to use this in the field.

SophisticaSean:
Super stoked to use this. Thank  you @danielbprice for your contribution here.

danielbprice:
@rs Any chance you would be willing to tag v1.30.0 so that people can start to pull this into projects more easily?

rs:
done

danielbprice:
Many thanks!

PR #560, opened 2023-07-04
feat: display error stack when using `.Fields()`

julienrbrt:
Display error stack when using `.Fields()`. Right now, even if the `ErrorStackMarshaler` is set, the stack will not be displayed if there is an error in the Fields() values but only when using Err().

### DIFF
--- a/fields.go
+++ b/fields.go
@@ -111,22 +111,6 @@ func appendFieldList(dst []byte, kvList []interface{}, stack bool) []byte {
 					dst = enc.AppendInterface(dst, m)
 				}
 
-				if stack && ErrorStackMarshaler != nil {
-					enc.AppendArrayDelim(dst)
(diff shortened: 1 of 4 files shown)


fields.go
```diff
@@ -74,6 +74,21 @@ func appendFieldList(dst []byte, kvList []interface{}) []byte {
 			default:
 				dst = enc.AppendInterface(dst, m)
 			}
+
+			if stack && ErrorStackMarshaler != nil {
+				dst = enc.AppendKey(dst, ErrorStackFieldName)
```
julienrbrt, line 79: Given that a `Fields()` can print multiple errors, should we append the key to `ErrorStackFieldName`?
If we do that, we could enable the display of _stacks_ for `[]error` as well.

julienrbrt:
Gentle bump @rs. Is that something that could be added?

PR #562, opened 2023-07-06
Introduce `AccessHandlerNew`

### LINKED ISSUE #561: Include bytes read in `hlog.AccessHandler` middleware
Currently, the `hlog.AccessHandler` middleware exposes the number of bytes written but does not surface the number of bytes read off of the request body.

As the introduction of a new field would break the `AccessHandler` API, I'd propose the creation of a new function that provides a struct to the callback function containing the number of read bytes.

```go
package hlog
//...
type AccessHandlerNewData struct {
  BytesRead int64
  BytesWritten int
  Status int
  Duration time.Duration
  Request *http.Request
}

func AccessHandlerNew(fn func(AccessHandlerNewData)) func(next http.Handler) http.Handler {}
//...
```

dillonstreator:
fixes #561


mitar:
This looks cool.

hlog/internal/mutil/body.go
```diff
@@ -0,0 +1,30 @@
...
+import "io"
+
+type byteCountReadCloser struct {
+	rc   io.ReadCloser
+	read int64
+}
+
+var _ io.ReadCloser = (*byteCountReadCloser)(nil)
+
+func NewByteCountReadCloser(body io.ReadCloser) *byteCountReadCloser {
```
mitar, line 12: In practice, http.body struct also implements `WriteTo`. So it might be reasonable to add counter here as well.

hlog/internal/mutil/body.go
```diff
@@ -0,0 +1,30 @@
...
+	return &byteCountReadCloser{
+		rc: body,
+	}
+}
+
+func (bcrc *byteCountReadCloser) Read(p []byte) (int, error) {
+	n, err := bcrc.rc.Read(p)
+	bcrc.read += int64(n)
```
mitar, line 20: I think those should be atomic operations.

hlog/internal/mutil/body.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (b *byteCountReadCloser) WriteTo(w io.Writer) (int64, error) {
+	n, err := io.Copy(w, b.rc)
+	atomic.AddInt64(&b.read, n)
+	return n, err
+}
+
+func (b *byteCountReadCloser) BytesRead() int64 {
+	return b.read
```
mitar, line 39: You should probably also here use atomic.

BTW, I [made it this way in my own code](https://gitlab.com/tozd/waf/-/blob/main/counters.go?ref_type=heads#L219-288).

hlog/hlog.go
```diff
@@ -216,3 +216,33 @@ func AccessHandler(f func(r *http.Request, status, size int, duration time.Durat
...
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			start := time.Now()
+
+			ww := mutil.WrapWriter(w)
+			body := mutil.NewByteCountReadCloser(r.Body)
+			r.Body = body
+
+			next.ServeHTTP(ww, r)
+			f(AccessHandlerData{
```
mitar, line 239: Also `f` should be called in defer. See current `AccessHandler` implementation.

PR #573, opened 2023-08-02
FilteredLevelWriter writes only logs at Level or above

### LINKED ISSUE #150: Multiple writers, different levels on each?
Hi,

I'm trying to configure zerolog to output logs both to the console and a file, with different formats. I can easily do that with `ConsoleWriter` and `io.MultiWriter`.

But I'm also trying to have the different log destinations log different levels.

I couldn't find a way do it. I tried to see if a Hook would solve it, but apparently the hooks seem to apply before the logs are sent to the Writers.

Is there a way to do it?

Thanks

mitar:
Fixes #150.

This allows one to combine `FilteredLevelWriter`, `LevelWriterAdapter` and `MultiLevelWriter` into any way you want. E.g., have info and above being logged to the console while all levels to a file.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -154,3 +154,29 @@ func ConsoleTestWriter(t TestingLog) func(w *ConsoleWriter) {
 		w.Out = TestWriter{T: t, Frame: 6}
 	}
 }
+
+// FilteredLevelWriter writes only logs at Level or above to Writer.
+//
+// It should be used only in combination with MultiLevelWriter when you
+// want to write to multiple destinations at different levels. Otherwise
+// you should just set the level on the logger and filter events early.
+// When using MultiLevelWriter then you set the level on the logger to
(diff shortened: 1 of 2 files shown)


mitar:
Maybe it should be named `FilteredLevelWriter`. I will push this right now.

rs:
The recommended way to do that is to set the level on the logger, so the event is skipped early.

mitar:
> The recommended way to do that is to set the level on the logger, so the event is skipped early.

Yes, but this is not possible if you want to log to multiple destinations (e.g., console and file). Then you have to set the level on the logger to the lowest of those you want and then filter at the writer.

rs:
Make sense. Can you please add a note on the description of this writer about the use-case it is covering so people don’t think that is the recommended way to deal with levels?

mitar:
Done!

PR #575, opened 2023-08-09
Add EmptyFields method to remove all the fileds from logger

GRbit:
Hello dear zerolog maintainers! First of all, thanks for the great logging package, which is also the fastest logger available. Hats off to everyone who worked on it.

This feature adds the ability to clear all previously added fields when creating a new logger.

Sometimes, when I pass the logger from one part of my service to another, I want to keep all the logger settings, but remove all the fields. For example, I can use the same field name to identify the context of the logger. If I reuse it, the result will have many fields with the same name, which doesn't look good:
`{"level":"info","context":"internal-logic","context":"db-adapter","context":"db-driver","message":"something"}`

Using this method, I can create a new logger instance like this:
`log.With().EmptyFields().Str("context","db-driver").Logger()`
And later have messages with only one "context" field:
`{"level":"info","context":"db-driver","message":"something"}`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,10 @@ _obj
 _test
 tmp
 
+# IDE config files
+.idea
+.vscode
+
 # Architecture specific extensions/prefixes
 *.[568vq]
 [568vq].out
--- a/.gitignore
+++ b/.gitignore
@@ -8,10 +8,6 @@ _obj
 _test
 tmp
 
-# IDE config files
-.idea
-.vscode
-
 # Architecture specific extensions/prefixes
 *.[568vq]
 [568vq].out
--- a/context.go
+++ b/context.go
@@ -384,7 +384,7 @@ func (c Context) Any(key string, i interface{}) Context {
 	return c.Interface(key, i)
 }
 
-// EmptyFields removes all the context fields.
+// Reset removes all the context fields.
 func (c Context) Reset() Context {
 	c.l.context = enc.AppendBeginMarker(make([]byte, 0, 500))
 	return c
--- a/context.go
+++ b/context.go
@@ -385,7 +385,7 @@ func (c Context) Any(key string, i interface{}) Context {
 }
 
 // EmptyFields removes all the context fields.
-func (c Context) EmptyFields() Context {
+func (c Context) Reset() Context {
 	c.l.context = enc.AppendBeginMarker(make([]byte, 0, 500))
 	return c
 }
--- a/context.go
+++ b/context.go
@@ -384,6 +384,12 @@ func (c Context) Any(key string, i interface{}) Context {
 	return c.Interface(key, i)
 }
 
+// EmptyFields removes all the context fields.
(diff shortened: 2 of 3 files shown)


rs:
What about renaming this method `Reset`?

madkins23:
Upvote from me, I have a use for this. I might suggest `ResetContext` as a rename but that's just overly picky. ;-)

.gitignore
```diff
@@ -8,6 +8,10 @@ _obj
 _test
 tmp
 
+# IDE config files
+.idea
+.vscode
+
```
rs, line 14:
```suggestion
```

rs: Unrelated

context.go
```diff
@@ -384,6 +384,12 @@ func (c Context) Any(key string, i interface{}) Context {
 	return c.Interface(key, i)
 }
 
+// EmptyFields removes all the context fields.
+func (c Context) EmptyFields() Context {
```
rs, line 388:
```suggestion
func (c Context) Reset() Context {
```

log_test.go
```diff
@@ -145,6 +145,25 @@ func TestWith(t *testing.T) {
 	}
 }
 
+func TestWithEmptyFields(t *testing.T) {
```
rs, line 148:
```suggestion
func TestWithReset(t *testing.T) {
```

log_test.go
```diff
@@ -145,6 +145,25 @@ func TestWith(t *testing.T) {
...
+	out := &bytes.Buffer{}
+	ctx := New(out).With().
+		Str("string", "foo").
+		Stringer("stringer", net.IP{127, 0, 0, 1}).
+		Stringer("stringer_nil", nil).
+		EmptyFields().
```
rs, line 154:
```suggestion
		Reset().
```

rs (changes requested):

.gitignore
madkins23: I have been guilty of this. Turns out there is a global `.gitignore` mechanism. Works like a charm, no need to set this kind of thing on a per project basis.

* https://gist.github.com/subfuzion/db7f57fff2fb6998a16c
* https://git-scm.com/docs/gitignore

GRbit: @madkins23 That's a neat thing to know! Thanks for sharing.

GRbit:
Thank you very much for the thorough review @rs !
I agree with all your suggestions, `Reset` indeed looks like a nice generic name (as well as `ResetContext`).

P.S.: it's very nice of you to highlight all the places where the function should be renamed, it just took me a few clicks.

context.go
```diff
@@ -384,6 +384,12 @@ func (c Context) Any(key string, i interface{}) Context {
 	return c.Interface(key, i)
 }
 
+// EmptyFields removes all the context fields.
```
tlipoca9, line 387:
```suggestion
// Reset removes all the context fields.
```

rs (approved):

PR #579, opened 2023-08-16
Missing trace output on syslog level writer

cringha:
Just like the title.  When i want to use the syslog level writer + syslog output send messages.  All others , e.g.  "debug", "info" .. worked well. But the trace func , i got empty from syslog service( using journald )  

I found there has a missing in syslog.go  WriteLevel .

I also created a issues https://github.com/rs/zerolog/issues/578 ,

### DIFF
--- a/syslog.go
+++ b/syslog.go
@@ -34,8 +34,8 @@ func SyslogLevelWriter(w SyslogWriter) LevelWriter {
 }
 
 // SyslogCEEWriter wraps a SyslogWriter with a SyslogLevelWriter that adds a
-// MITRE CEE prefix for JSON syslog entries, compatible with rsyslog 
-// and syslog-ng JSON logging support. 
+// MITRE CEE prefix for JSON syslog entries, compatible with rsyslog
+// and syslog-ng JSON logging support.
 // See https://www.rsyslog.com/json-elasticsearch/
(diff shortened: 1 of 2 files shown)


cringha:
The code like this 
`
	logwriter, _ := syslog.New(syslog.LOG_DEBUG, "demo")
	ll := zerolog.New(zerolog.SyslogLevelWriter(logwriter))
	level := "trace" 
	
	zl, err := zerolog.ParseLevel(level)
	if err != nil {
		fmt.Printf("Error parse level %v\n", err)
		return
	}

	fmt.Printf("test %s \n", level)
	ll.WithLevel(zl).Msgf("Message: level %s %v", level, time.Now())`

cringha:
Fixed, some error in ut

mitar:
I think this was done on purpose. Mapping zerolog trace to syslog debug might be surprising to people. Syslog does not have trace equivalent.

I think this could be instead better done with some custom LevelWriter which would map trace level to debug level and then you pass that on to syslog writer. Then it is explicit what you want to do. It is then just a question of your configuration. I think existing zerolog syslog implementation is fine as is.

mitar:
I realized that it is not really possible to map level inside a writer because you would also have to change `level` field inside the JSON payload so you would have to parse/marshal JSON again. Level mapping should be done at some other level it seems.

PR #580, opened 2023-08-18
chore: adding any function to context

### LINKED ISSUE #576: Add context function for any
similar to https://github.com/rs/zerolog/issues/514 
now here is `Any` in event, need a similar change in context, so `log.With()` can use any 

```
// before
log.With().Interface("params", p).Msg("charge request")

// after
log.With().Any("params", p).Msg("charge request")

```

shettyh:
fix #576

### DIFF
--- a/context.go
+++ b/context.go
@@ -381,8 +381,7 @@ func (c Context) Interface(key string, i interface{}) Context {
 
 // Any is a wrapper around Context.Interface.
 func (c Context) Any(key string, i interface{}) Context {
-	c.l.context = enc.AppendInterface(enc.AppendKey(c.l.context, key), i)
-	return c
+	return c.Interface(key, i)
 }
 
 type callerHook struct {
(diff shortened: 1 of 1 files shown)


context.go
```diff
@@ -379,6 +379,12 @@ func (c Context) Interface(key string, i interface{}) Context {
 	return c
 }
 
+// Any is a wrapper around Context.Interface.
+func (c Context) Any(key string, i interface{}) Context {
+	c.l.context = enc.AppendInterface(enc.AppendKey(c.l.context, key), i)
```
mitar, line 384: Comment is saying that this is a wrapper around `Context.Interface` but you copied its implementation and you are not calling into `Context.Interface`? I think you can just call into `Context.Interface`? Compiler will probably inline this anyway.

shettyh: Sorry, fixed it

PR #586, opened 2023-09-04
added support for NO_COLOR

### LINKED ISSUE #582: Add support for `NO_COLOR`
Support should be added for [`NO_COLOR`](https://no-color.org)

Gaardsholt:
This should fix #582

### DIFF
--- a/console.go
+++ b/console.go
@@ -312,6 +312,11 @@ func needsQuote(s string) bool {
 
 // colorize returns the string s wrapped in ANSI code c, unless disabled is true.
 func colorize(s interface{}, c int, disabled bool) string {
+	e := os.Getenv("NO_COLOR")
+	if e == "true" {
+		disabled = true
+	}
+
 	if disabled {
 		return fmt.Sprintf("%s", s)
 	}


console.go
```diff
@@ -312,6 +312,11 @@ func needsQuote(s string) bool {
 
 // colorize returns the string s wrapped in ANSI code c, unless disabled is true.
 func colorize(s interface{}, c int, disabled bool) string {
+	e := os.Getenv("NO_COLOR")
+	if e == "true" {
```
rs, line 316: [GNU libtextstyle](https://www.gnu.org/software/gettext/libtextstyle/manual/html_node/The-NO_005fCOLOR-variable.html) doc says:
> The environment variable NO_COLOR can be used to suppress styling in the textual output. When this environment variable is set (to any value), libtextstyle-enabled programs will not emit colors and other text styling.

rs (changes requested):

console.go
Gaardsholt: Ahh you are right, fixed the code now.

rs (approved):

PR #593, opened 2023-10-07
Update README.md

ShiJieCloud:
default log level is trace , not debug

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ func main() {
 // Output: {"time":1516134303,"level":"debug","message":"hello world"}
 ```
 > Note: By default log writes to `os.Stderr`
-> Note: The default log level for `log.Print` is *debug*
+> Note: The default log level for `log.Print` is *trace*
 
 ### Contextual Logging
 


ShiJieCloud:
from code:
`
// Logger is the global logger.
var Logger = zerolog.New(os.Stderr).With().Timestamp().Logger()

func New(w io.Writer) Logger {
	if w == nil {
		w = ioutil.Discard
	}
	lw, ok := w.(LevelWriter)
	if !ok {
		lw = LevelWriterAdapter{w}
	}
	return Logger{w: lw, level: TraceLevel}
}`

ShiJieCloud:
from code:
```go
// Logger is the global logger.
var Logger = zerolog.New(os.Stderr).With().Timestamp().Logger()
```
```go
func New(w io.Writer) Logger {
if w == nil {
w = ioutil.Discard
}
lw, ok := w.(LevelWriter)
if !ok {
lw = LevelWriterAdapter{w}
}
return Logger{w: lw, level: TraceLevel}
}```

PR #594, opened 2023-10-09
Use `sync.Pool` and `bytes.Buffer` for `Logger.With` perf improvement

a-menshchikov:
There are currently a lot of allocations in the `Logger.With` due to direct `make` calls. This PR replaces it by using `sync.Pool` and `bytes.Buffer`.

### DIFF
--- a/log.go
+++ b/log.go
@@ -114,6 +114,7 @@
 package zerolog
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
@@ -122,7 +123,14 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 )
+
+var contextBufPool = &sync.Pool{
+	New: func() interface{} {
+		return bytes.NewBuffer(make([]byte, 0, 500))
+	},
+}
 
 // Level defines log levels.
 type Level int8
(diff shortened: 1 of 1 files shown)


rs:
Thanks, I'll check the PR. Just curious, why are you calling With so often? The goal of With is to call it once for a logger.

a-menshchikov:
@rs I have the helper function that gets a logger from context and adds some trace attributes (trace id and span id) on each call. It brings some overhead, but looks good for me and allows my team to write simpler code.

P.S. I see data races, now I try to solve this issue.

a-menshchikov:
@rs it seems I don't know how to fix data races, which occured by simultaneously use of `-race` and `-bench`. =(

Could you suggest some solution?

rs:
Your implementation is actually not correct as the buffer is used after you put it back in the pool. If you want to pull the buffer, you need to pull it in `With` and put it back once the logger is no longer used. With the current API, there is no way you could implement such pooling.

a-menshchikov:
@rs thank you for clarifying, you are absolutely right, it's my fail. I have tried several ways to work around this problem, but it seems like I need to use `Event.Fields` instead of `Logger.With` to solve my problem.

PR #596, opened 2023-10-13
Update example image to how it is currently rendered

mitar:
I noticed that the image is not how currently default console writer renders it.

The image was done using Ubuntu 20.04 using Gnome Terminal (I think default color settings).

I also made a JSONL file with log entries to be able to generate this in the future using the `prettylog` tool.

BTW, prettylog tool does not work on files, see [my comment here](https://github.com/rs/zerolog/pull/449#discussion_r1358116485).


rs:
I added support for file input in 6e8841b038192f6cda1bf2652418cabc8928289c

mitar:
One question. I noticed this because I wanted to compare how my changed colors compare to the default colors. I think my changed colors are much more readable. I wonder if this is something you would entertain that I make a PR and change also for defaults in zerolog:

Screenshot from 2023-10-13 14-12-06

Changes:
* Levels do not have bold to not draw focus.
* Debug level is not yellow but gray, while warning is yellow (it was really confusing to me).
* Messages are bold (to draw focus) at info level and above.
* Error message are bold (to draw focus).

This makes debug and trace messages not stand out (they are all gray) while important ones do (especially important because there are generally many more debug ones).

rs:
sure go ahead

PR #602, opened 2023-10-15
Add TriggerLevelWriter

### LINKED ISSUE #583: Support canonical log lines & log level per request
## Canonical log lines

There are generally two ways to organize log lines which are all related to a particular request:
* Generate some type of request ID and then include that request ID in all log lines related with the request.
* Issue only one log line for the whole request, accumulating all logging into it. This can be called [canonical log lines](https://brandur.org/canonical-log-lines).

In a way the this is standard question of normalized vs. denormalized data.

There are some existing issues I could find which seems to talk about support for canonical log lines in zerolog: #362 and #408, but I think it is worth investigating if zerolog should&could support canonical log lines out of the box.

The support for request ID is easy with sub-loggers. And I think it could be the same for canonical log lines. In fact, it could also be organized as sub-loggers, maybe only with some additional `.Nested()` or some similar call, which would turn on canonical log lines feature. Then, when a sub-logger would be called to send the event, instead of writing it out to the writer, the log line would accumulate inside parent's logger array field associated with the level. Only when the parent logger then issues the event, the log line would be printed out. So the idea here is:

* At the beginning of the request handling you obtain a canonical log lines logger (e.g., call `.Nested()` or something).
* You pass this around inside the context, everythhing logs using it.
* After handling finishes, you call `.Send()` or something, maybe even adding more metadata (like amount of data transferred inside the request handling, etc.).

The structure could be something like:

```yaml
{
  "warn": [...],
  "error": [...],
}
```

And so on for different levels.

## Log level per request

An interesting thing with canonical log lines is also how level filtering works. I propose that the canonical log line gets a level of the most severe event level logged into it. For example, if both error and warn levels were logged for the same request, then the canonical log level gets the error level.

With a canonical log lines, the sub-logger could log at a very detailed logging level (e.g., debug) but the whole line could be issues only if there was an error logged as well by setting minimal log level to error. This would allow developers to have debugging information together with all errors, but only with errors.

Such type of logging is often called ["log level per request"](https://devopedia.org/log-level-per-request). To my understanding zerolog does not currently support that either. This could be supported for both canonical log lines and for regular log lines. This is where we can see that this is not just a question of de/normalization of log lines, but about processing of log events in batches.

## Batching log events

Both canonical log lines and log level per request show a need for some "batching of log events" feature in zerolog. Where log events would be batched together until some later event decides to write them out or not (potentially modifying them). In the case of canonical log lines, batching would batch all log events for a sub-logger and then write them out as one log event. In the case of log level per request, after batch finishes, logic would decide to write it out or not.

(I am not sure if "batching" is the right term here. Maybe "batch processing" or something?)

## Alternatives

Alternatives to both are possible by post-processing logs:
* You can store all logs into a searchable index and then group log entries by request ID.
* You could always log at debug level and then remove debugging log entries later on from the index if they are not associated with errors. This is not very performant though and generally logging indices behave well when they are append only.
* You could imagine a zerolog writer which does such combining of log entries by request ID into canonical log lines. This would require paring JSON back, combining, and then writing new JSON out, again not very performant.
* Similarly a zerolog writer could filter log entries based on presence of errors per request. But again that requires JSON parsing inside the writer.

mitar:
Fixes #583.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -180,3 +180,135 @@ func (w *FilteredLevelWriter) WriteLevel(level Level, p []byte) (int, error) {
 	}
 	return len(p), nil
 }
+
+var triggerWriterPool = &sync.Pool{
+	New: func() interface{} {
+		return bytes.NewBuffer(make([]byte, 0, 1024))
+	},
+}
+
+// TriggerLevelWriter buffers log lines at the ConditionalLevel or below
+// until a trigger level (or higher) line is emitted. Log lines with level
+// higher than ConditionalLevel are always written out to the destination
+// writer. If trigger never happens, buffered log lines are never written out.
+//
+// It can be used to configure "log level per request".
+type TriggerLevelWriter struct {
+	// Destination writer. If LevelWriter is provided (usually), its WriteLevel is used
+	// instead of Write.
+	io.Writer
+
+	// ConditionalLevel is the level (and below) at which lines are buffered until
+	// a trigger level (or higher) line is emitted. Usually this is set to DebugLevel.
+	ConditionalLevel Level
+
+	// TriggerLevel is the lowest level that triggers the sending of the conditional
+	// level lines. Usually this is set to ErrorLevel.
+	TriggerLevel Level
+
+	buf       *bytes.Buffer
+	triggered bool
+	mu        sync.Mutex
+}
+
+func (w *TriggerLevelWriter) WriteLevel(l Level, p []byte) (n int, err error) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	// At first trigger level or above log line, we flush the buffer and change the
+	// trigger state to triggered.
+	if !w.triggered && l >= w.TriggerLevel {
+		err := w.trigger()
+		if err != nil {
+			return 0, err
+		}
+	}
+
+	// Unless triggered, we buffer everything at and below ConditionalLevel.
+	if !w.triggered && l <= w.ConditionalLevel {
+		if w.buf == nil {
+			w.buf = triggerWriterPool.Get().(*bytes.Buffer)
+		}
+
+		// We prefix each log line with a byte with the level.
+		// Hopefully we will never have a level value which equals a newline
+		// (which could interfere with reconstruction of log lines in the trigger method).
+		w.buf.WriteByte(byte(l))
+		w.buf.Write(p)
+		return len(p), nil
+	}
+
+	// Anything above ConditionalLevel is always passed through.
+	// Once triggered, everything is passed through.
+	if lw, ok := w.Writer.(LevelWriter); ok {
+		return lw.WriteLevel(l, p)
+	}
+	return w.Write(p)
+}
+
+// trigger expects lock to be held.
+func (w *TriggerLevelWriter) trigger() error {
+	if w.triggered {
+		return nil
+	}
+	w.triggered = true
+
+	if w.buf == nil {
+		return nil
+	}
+
+	p := w.buf.Bytes()
+	for len(p) > 0 {
+		// We do not use bufio.Scanner here because we already have full buffer
+		// in the memory and we do not want extra copying from the buffer to
+		// scanner's token slice, nor we want to hit scanner's token size limit,
+		// and we also want to preserve newlines.
+		i := bytes.IndexByte(p, '\n')
+		line := p[0 : i+1]
+		p = p[i+1:]
+		// We prefixed each log line with a byte with the level.
+		level := Level(line[0])
+		line = line[1:]
+		var err error
+		if lw, ok := w.Writer.(LevelWriter); ok {
+			_, err = lw.WriteLevel(level, line)
+		} else {
+			_, err = w.Write(line)
+		}
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// Trigger forces flushing the buffer and change the trigger state to
+// triggered, if the writer has not already been triggered before.
+func (w *TriggerLevelWriter) Trigger() error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	return w.trigger()
+}
+
+// Close closes the writer and returns the buffer to the pool.
+func (w *TriggerLevelWriter) Close() error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.buf == nil {
+		return nil
+	}
+
+	// We return the buffer only if it has not grown above the limit.
+	// This prevents accumulation of large buffers in the pool just
+	// because occasionally a large buffer might be needed.
+	if w.buf.Cap() <= TriggerLevelWriterBufferReuseLimit {
+		w.buf.Reset()
+		triggerWriterPool.Put(w.buf)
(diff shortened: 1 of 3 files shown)


mitar:
In my own code, I then made:

```go
func WithContext(ctx context.Context, writer io.Writer, level, conditionalLevel, triggerLevel zerolog.Level) (context.Context, func(), func()) {
	w := zerolog.NewTriggerLevelWriter(
		writer,
		conditionalLevel,
		triggerLevel,
	)
	ctxLogger := zerolog.New(w).Level(level).With().Timestamp().Logger()
	closeCtx := func() {
		_ = w.Close()
	}
	trigger := func() {
		_ = w.Trigger()
	}
	return ctxLogger.WithContext(ctx), closeCtx, trigger
}
```

One has then to call `closeCtx` after the context is not used anymore.

writer.go
```diff
@@ -180,3 +181,145 @@ func (w *FilteredLevelWriter) WriteLevel(level Level, p []byte) (int, error) {
...
+func (w *TriggerLevelWriter) Close() error {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.buf == nil {
+		return nil
+	}
+
+	w.buf.Reset()
+	triggerWriterPool.Put(w.buf)
```
rs, line 321: Should we add a size limit for returning the buffer to the pool? We may end up with some large buffer containing something like a stack trace we don't want to maintain in memory via the pool.

mitar: I would guess that if one log line contains a stack trace, probably also other log lines from the same program contains them. So it is probably OK to have all buffers around as they are and grow to. Otherwise buffers will have to grow again and again to that size?

We could also make a setting in the writer field, for like max capacity to return to the pool with 0 (default) being that it returns everything.

mitar: If you set that field, do we resize back to max capacity buffers before returning them to the pool (instead of dropping them completely)?

I would vote for keeping it as-is though.

rs: Imagine 1% of logs triggering a larger amount of logs, they will over time (pretty quickly actually) "pollute" the whole pool with large buffers and waste memory.

Those large buffers needs to be dropped, resizing won't change the memory taken by the backing array.

mitar: So I add an option or we hard-code some value?

mitar: OK, thank you for pointing this out. I read more about this now and it does not seem to be clear solution.

Related Golang issues:
* https://github.com/golang/go/issues/23199 (general issue that sync.Pool seems to be primarily suitable when stored data is of similar size)
* https://github.com/golang/go/issues/27735 (JSON encoding has a similar issue)

Some solutions seems to be:
* having a hard-coded limit as you proposed:
  * fmt standard library [has 64 KB limit](https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/fmt/print.go;l=169) and it seems it works well enough because most string formatting is on short strings
  * but trying to do a [similar limit for JSON encoding](https://go-review.googlesource.com/c/go/+/136035) had a large negative impact on performance and was abandoned and is [not yet solved](https://github.com/golang/go/issues/27735 (JSON) - I think our use case is closer to JSON encoding and less of fmt formatting
* do some local (per buffer) or global (per pool) statistics and decide which buffers to drop
  * https://github.com/valyala/bytebufferpool seems to do something like that, maybe there are also other 3rd party libraries to use here instead of trying to fix it ourselves
  * [this is an example of doing some local stats](https://github.com/golang/go/issues/27735#issuecomment-739169121) proposed for JSON encoding, but it seems it has not yet been found as the best approach
* do buffer segmentation (have different pools for different sizes of buffers) - not suitable for us because we do not know in advance the size we need

nhooyr: You could implement `.Available()` with the API prior to Go 1.21. It's just cap minus len.

mitar: I do not think you can, [looking at the code](https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/bytes/buffer.go;l=77-86) because you do not have `b.off`.

nhooyr: Oh right. I have an intuition that grabbing buf.Length and buf.Capacity before buf.Reset and then grabbing them again after should give you enough information to figure out buf.Available but I'm not 100% sure rn.

mitar: @rs So this is waiting for your input about the direction to take this.

rs: I would keep it simple and use a sensible limit.

mitar: @rs: I have updated PR. It now drops the buffer if it grows too much. Default is 64 KB, based on fmt limit. But it can be changed if somebody wants/needs.

writer.go
```diff
@@ -180,3 +181,150 @@ func (w *FilteredLevelWriter) WriteLevel(level Level, p []byte) (int, error) {
 	}
 	return len(p), nil
 }
+
+var triggerWriterPool = &sync.Pool{
+	New: func() interface{} {
+		return new(bytes.Buffer)
```
rs, line 187: What about initializing the buffer with a minimum size to avoid initial allocations?

writer.go
```diff
@@ -180,3 +181,150 @@ func (w *FilteredLevelWriter) WriteLevel(level Level, p []byte) (int, error) {
...
+// NewTriggerLevelWriter returns a new TriggerLevelWriter.
+//
+// It obtains a buffer from the pool and you must
+// call Close to return the buffer to the pool.
+func NewTriggerLevelWriter(w io.Writer, conditionalLevel, triggerLevel Level) *TriggerLevelWriter {
+	return &TriggerLevelWriter{
+		Writer:           w,
+		ConditionalLevel: conditionalLevel,
+		TriggerLevel:     triggerLevel,
+		buf:              triggerWriterPool.Get().(*bytes.Buffer),
```
rs, line 200: Extra optimization could be to get from the pool on first write instead of unconditionally.

writer.go
```diff
@@ -180,3 +181,150 @@ func (w *FilteredLevelWriter) WriteLevel(level Level, p []byte) (int, error) {
...
+	return w.Write(p)
+}
+
+// trigger expects lock to be held.
+func (w *TriggerLevelWriter) trigger() error {
+	if w.triggered {
+		return nil
+	}
+	w.triggered = true
+	defer w.buf.Reset()
```
rs, line 270: Reset is called on pool put already

mitar:
@rs: I addressed your review comments.

writer.go
```diff
@@ -180,3 +181,135 @@ func (w *FilteredLevelWriter) WriteLevel(level Level, p []byte) (int, error) {
...
+	triggered bool
+	mu        sync.Mutex
+}
+
+func (w *TriggerLevelWriter) WriteLevel(l Level, p []byte) (n int, err error) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.buf == nil {
+		w.buf = triggerWriterPool.Get().(*bytes.Buffer)
```
rs, line 220: This should be inside the `if !w.triggered && l <= w.ConditionalLevel` condition below and `trigger()` should be modified to accept a `nil` buffer.

mitar: Done.

PR #607, opened 2023-10-21
Add additional hlog logging handlers

mitar:
* `HTTPVersionHandler`: Similar to `ProtoHandler`, just without `HTTP/` prefix. In practice, there is always `HTTP/` prefix, so we can drop that (if it exists, which it does) to have more compact logs.
* `RemoteAddrHandlerWithoutPort`: Similar to `RemoteAddrHandler`, but do you really care about a remote port? Remote port makes it hard to aggregate logs based on remote IP.
* `HostHandlerWithoutPort`: Is similar, generally you do not care on which local port you run stuff.
* `EtagHandler`: Logs response header, `Etag`, can help with debugging which version of content user got.
* `ResponseHeaderHandler`: Similar to `CustomHeaderHandler`, but for response headers. If you do canonical log lines, you probably have middleware which logs it after everything is handled, so this handler should be earlier in the middleware stack.

### DIFF
--- a/hlog/hlog.go
+++ b/hlog/hlog.go
@@ -271,6 +271,23 @@ func EtagHandler(fieldKey string) func(next http.Handler) http.Handler {
 	}
 }
 
+func ResponseHeaderHandler(fieldKey, headerName string) func(next http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			defer func() {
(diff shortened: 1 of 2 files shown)


rs:
* `RemoteAddrHandlerWithoutPort` would probably be better named `RemoteIPHandler`
* `HostHandlerWithoutPort` could just be a `HostHandler` with an optional (variadic) bool parameter `trimPort`.

mitar:
Renamed.

PR #611, opened 2023-10-30
hlog: Don't write StatusOK header for websockets

rschmied:
When logging requests and the request is a websocket and the websocket is not properly closed (e.g. the client reloads the page, disconnecting the socket), then hlog tries to write `http.StatusOK` via `WriteHeader()` to the hijacked connection which fails as described here

https://stackoverflow.com/questions/32657603/why-do-i-get-the-error-message-http-response-write-on-hijacked-connection

This PR addresses this problem by passing a flag into the `WrapWriter()` func to indicate whether the request is a Websocket. In this case, the `WriteHeader()` is suppressed.

### DIFF
--- a/hlog/internal/mutil/writer_proxy.go
+++ b/hlog/internal/mutil/writer_proxy.go
@@ -29,13 +29,13 @@ type WriterProxy interface {
 
 // WrapWriter wraps an http.ResponseWriter, returning a proxy that allows you to
 // hook into various parts of the response process.
-func WrapWriter(w http.ResponseWriter) WriterProxy {
+func WrapWriter(w http.ResponseWriter, isWs bool) WriterProxy {
 	_, cn := w.(http.CloseNotifier)
 	_, fl := w.(http.Flusher)
 	_, hj := w.(http.Hijacker)
 	_, rf := w.(io.ReaderFrom)
 
-	bw := basicWriter{ResponseWriter: w}
+	bw := basicWriter{ResponseWriter: w, isWs: isWs}
 	if cn && fl && hj && rf {
 		return &fancyWriter{bw}
 	}
@@ -49,6 +49,7 @@ func WrapWriter(w http.ResponseWriter) WriterProxy {
 // http.ResponseWriter interface.
 type basicWriter struct {
 	http.ResponseWriter
+	isWs        bool
 	wroteHeader bool
 	code        int
 	bytes       int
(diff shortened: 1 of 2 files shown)


hlog/internal/mutil/writer_proxy.go
```diff
@@ -59,7 +60,11 @@ func (b *basicWriter) WriteHeader(code int) {
 	if !b.wroteHeader {
 		b.code = code
 		b.wroteHeader = true
-		b.ResponseWriter.WriteHeader(code)
+		// if it's a websocket then we can't actually write the header
+		// on a hijacked session
+		if !b.isWs {
```
mitar, line 65: Who exactly calls `WriteHeader` in question? Because probably this check should be made there and not inside `WriteHeader`.

Also, for hijacked connections `AccessHandler` has little sense and will not really count bytes going over the connection.

rschmied: Not sure if I follow? `hlog` wraps the original request, when the request is done, `hlog` closes the wrapped request/response and sets the status. It also tries to write a header with this status into the response. But this yields an error if the response has been hijacked (e.g. it is a websocket).

mitar: I am trying to understand where this happens.

> closes the wrapped request/response

Where does it do that?

> It also tries to write a header with this status into the response.

Can you share a stack trace at the point this happens?

rschmied: You are right. I looked at this again, the error is actually within Gorilla websocket (e.g. https://github.com/gorilla/websocket/issues/852) -- not sure why I did see a (temporary) remediation of the issue by applying this patch. But it has shown again and it's not zerolog. It shows b/c I added the std log handler to zerolog in my program. Closing the PR. Thanks for pointing it out.

rschmied:
closing as not applicatble.

hlog/internal/mutil/writer_proxy.go
mitar: I have recently used https://github.com/nhooyr/websocket in a stack with zerolog and it worked well for me.

rschmied: Thanks for mentioning this. I am using gorilla websockets quite extensively but maybe this is a worthy alternative. I was quite happy that they restarted the Gorilla suite eventually but with those latest changes things have gotten worse, not better (at least the parts that I am observing / need)

PR #618, opened 2023-11-17
sampling: Support message-level sampling

### LINKED ISSUE #543: Sampled logs used in multiple call point the sampling bucket
```
sampled := log.Sample(&zerolog.BasicSampler{N: 100})

sampled.Info().Msg("call point1")
...
sampled.Info().Msg("call point2")
```
In this example both call points will share the sampling bucket and for example call point1 is a lot more verbose, there is high probability call point2 never gets logged as it has less probability of getting sampled. One approach is to declare a separate logger for each call point but this gets unwieldy, especially when we can't declare them close to the call point, for we don't want to instantiate the logger inside the loop as sampler will get reset on each loop iteration.

another approach make sampler call point aware, and use runtime.Caller() to determine where it's being called from and instantiate separate bucket for separate call points, however runtime.Caller() is pretty expensive and we don't want to call it on each log ?

last approach is to use some sort of go:generate magic that instantiate unique vars for each call point, however it's pretty tricky to get this one right

So my question, has zerolog community come across this problem and how did you solve this ?

abhinav:
zerolog's Sampler interface is currently limited
to sampling on the log level.
This makes it difficult to build a smarter sampler
that suppresses repetitive log messages
while still surfacing unique messages.
For comparison, Zap uses `(level, msg)` as the sampling key[^1],
which allows it to do the above.

[^1]: https://github.com/uber-go/zap/blob/v1.24.0/zapcore/sampler.go#L221

This change proposes a new interface, MessageSampler:

    type MessageSampler {
        SampleMessage(Level, string) bool
    }

There are two ways for us to incorporate this interface:
optional upgrade or independent concept.

For convenience, I've included both versions as separate commits.

### Optional upgrade

If the sampler passed to `Logger.Sample` implements this interface,
the SampleMessage method will be used by `*Event`
when it finally has access to the message content.

This turns the sampler interface into a two-stage thing.
We log only if `Sample(level) && SampleMessage(level, msg) == true`.
There's risk of confusion where a SampleMessage implementation
calls Sample again--double counting that message.

### Independent concept

We treat `MessageSampler` as an independent concept.
You can register a Sampler, a MessageSampler, neither, or both.

We still log only if `Sample(level) && SampleMessage(level, msg) == true`
but this time there's lower risk that a SampleMessage implementation
will call Sample because there's no expectation to satisfy both interfaces.

In exchange, this needs a new API:

    func (Logger) SampleMessages(MessageSamler) Logger

---

As I said, this PR contains both versions as separate commits.
I think *Independent concept* might make more sense,
but I'm interested in hearing which approach the maintainers prefer.

Resolves #543

### DIFF
--- a/event.go
+++ b/event.go
@@ -29,6 +29,7 @@ type Event struct {
 	ch        []Hook          // hooks from context
 	skipFrame int             // The number of additional frames to skip when printing the caller.
 	ctx       context.Context // Optional Go context for event
+	sampler   MessageSampler  // optional: if set, used to sample messages
 }
 
 func putEvent(e *Event) {
@@ -142,6 +143,11 @@ func (e *Event) msg(msg string) {
 	for _, hook := range e.ch {
 		hook.Run(e, e.level, msg)
 	}
+
+	if e.sampler != nil && !e.sampler.SampleMessage(e.level, msg) {
+		return
+	}
+
 	if msg != "" {
 		e.buf = enc.AppendString(enc.AppendKey(e.buf, MessageFieldName), msg)
 	}
--- a/event.go
+++ b/event.go
@@ -21,15 +21,15 @@ var eventPool = &sync.Pool{
 // Event represents a log event. It is instanced by one of the level method of
 // Logger and finalized by the Msg or Msgf method.
 type Event struct {
-	buf       []byte
-	w         LevelWriter
-	level     Level
-	done      func(msg string)
-	stack     bool            // enable error stack trace
-	ch        []Hook          // hooks from context
-	skipFrame int             // The number of additional frames to skip when printing the caller.
-	ctx       context.Context // Optional Go context for event
-	sampler   MessageSampler  // optional: if set, used to sample messages
+	buf        []byte
+	w          LevelWriter
+	level      Level
+	done       func(msg string)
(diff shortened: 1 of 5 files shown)


abhinav:
Gentle bump, @rs.
Are y'all willing to accept something in this direction?

rs:
Suppressing duplicates messages is not the same as sampling, it is a different concept. I wonder if we could add a a method to tell an event not to be sent from a Hook in order to handle such case instead.

abhinav:
> Suppressing duplicates messages is not the same as sampling, it is a different concept.

The idea was not to just suppress all duplicate messages, but to still let-through a subset of the duplicates—sampled using the `(level, message)` tuple to create buckets. Although, I agree, practically, this would suppress a number of duplicate messages.

> I wonder if we could add a a method to tell an event not to be sent from a Hook in order to handle such case instead.

That's a valid approach too!
Would the existing `(*Event).Discard` method work here?

rs:
Oh I forgot about the `Discard` method, so yes it is already possible with no change.

abhinav:
Ack. We can build some sampling logic around that.
Closing the PR—unless you're interested in message-granularity sampling.
Thanks for your response!

PR #623, opened 2023-12-07
add: ContextLogger Interface LogObjectMarshaler

Hyuga-Tsukui:
The issue I've identified is that the `Interface` method, derived from `Ctx`, does not support `LogObjectMarshaler`. 

I made a PR out of the blue because it is expected to be minor, do I need an ISSUE?

PR #631, opened 2024-01-08
Fixed failing Tests #623

Hyuga-Tsukui:
I am not sure if this is the right response.

Originally, there was no detailed test as I wrote when feature https://github.com/rs/zerolog/commit/fdbdb0963042e1b3e95bb1fae826dfd92a1c4bc3 was added.
I believe this is because it is fully tested in each Log test.

The reason I added the test this time was to easily check the behavior (Example may be sufficient)


Hyuga-Tsukui:
This PR is associated with  #623

rs:
Fixed by another PR

PR #635, opened 2024-01-13
Decode CBOR in the diode test to fix workflow failures

gmbuell:


gmbuell:
Merged as part of #636

PR #645, opened 2024-02-03
Add prettylog time format

freak12techno:
This PR adds a way to specify time format for logs displaying via --time-format flag.
Motivation for doing so: sometimes I need to look at really big amounts of logs, and it's not clear from time.Kitchen format which day these logs are from and which timezone they are in, so I want to be able to see this as well.

This adds importing of the 'flags' package, but this package is built in so this shouldn't cause any issues (my goals were to avoid adding something as a dependency, like cobra/viper).

Also this adds a base for future logging customisation via flags.

### DIFF
--- a/cmd/prettylog/prettylog.go
+++ b/cmd/prettylog/prettylog.go
@@ -3,9 +3,11 @@ package main
 import (
 	"bufio"
 	"errors"
+	"flag"
 	"fmt"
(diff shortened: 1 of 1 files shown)


freak12techno:
@rs is it possible to also tag a new release with this one, just like you did with my previous PR?

Issue #571, opened 2023-07-31
stdlib slog handler for zerolog


shettyh:
Hi,

Is there any plan to add [slog](https://pkg.go.dev/golang.org/x/exp/slog) handler implementation for Zerolog ?. If there are no ongoing efforts then i can raise the PR the same

Regards

rs:
No plan at the moment. What would be the use-case?

euskadi31:
To no longer depend directly on a log library?

rs:
Isn't slog self sufficient?

shettyh:
@rs thanks for the response, `slog` provides default implementation for Json and Text but those are not as performant as zerolog. So having the handler interface implemented for zerolog will help to migrate to slog and no longer need to depend on specific logging library in application code

rs:
Let’s try it. The risk is that the difference in performance comes from the API model. Please make sure to bench it so we don’t make false promises :)

jba:
Slog author here. Thanks @shettyh for beating me here! Slog will be in go 1.21, so it makes sense to provide some integration with it. I was going to suggest a slog.Handler that writes to a zerolog logger. @phsym nails the use case: letting packages that use slog and ones that use zerolog produce consistent output.

phsym:
There will be a performance cost anyway.

Benchmarking slog with a "no-op" backend vs zerolog shows that slog frontend already brings some overhead. But still it will be great to have a zerolog handler for the libraries which will use slog.

No-Op handler:
```go
type DummyHandler struct{}

func (*DummyHandler) Enabled(context.Context, slog.Level) bool {
	return true
}
func (*DummyHandler) Handle(context.Context, slog.Record) error {
	return nil
}
func (h *DummyHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
	return h
}
func (h *DummyHandler) WithGroup(name string) slog.Handler {
	return h
}
```

Bench:
```go
func BenchmarkDummy(b *testing.B) {
	ctx := context.Background()
	l := slog.New(&DummyHandler{})
	l = l.With("foo", "bar")
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		l.LogAttrs(ctx, slog.LevelInfo, "hello", slog.String("bar", "baz"))
	}
}

func BenchmarkZerolog(b *testing.B) {
	l := zerolog.New(io.Discard).Level(zerolog.DebugLevel).With().Timestamp().Logger()
	l = l.With().Str("foo", "bar").Logger()
[...]

PR #653, opened 2024-02-29
Fix phsym/zeroslog breakage due to rs/zerolog v1.32.0

madkins23:
TL;DR Add function to create a zerolog.Context object with a logger copy and clear the copy's context byte array in order to fix breakage in `phsym/zeroslog`.

The recent release of `rs/zerolog v1.32.0` broke `phsym/zeroslog`. The latter was creating blank `zerolog.Context` objects in several places and they worked fine up through `rs/zerolog v1.31.0`. I documented the issue in a [`phsym/zeroslog` PR](https://github.com/phsym/zeroslog/pull/4), wherein I provided a potential solution to the issue that didn't require any changes to `rs/zerolog`.

This PR provides a (hopefully) minor change to `rs/zerolog` that makes fixing `phsym/zeroslog` simpler. I have already filed a second presumptive [`phsym/zeroslog` PR](https://github.com/phsym/zeroslog/pull/6) dependent on the changes in this `rs/zerolog` PR.

Hopefully this PR will be an acceptable and minor change.

I gave the new function a long name because it is combining two issues. It could be broken down into:

* `NewContext(logger Logger)`  
  which would just create the new `zerolog.Context` object with a logger and
* `(l Logger) ClearContext()`  
  which would clear the `context` byte array on the copied `Logger`.

These two calls could be used instead of the single `NewContextWithResetLogger` call, something like this:
```Go
newCtx := NewContext(logger)
newCtx.Logger().ClearContext()
```

If the latter is preferable I can generate a different PR.

### DIFF
--- a/context.go
+++ b/context.go
@@ -14,6 +14,11 @@ type Context struct {
 	l Logger
 }
 
+func NewContextWithResetLogger(l Logger) Context {
+	l.context = enc.AppendBeginMarker(make([]byte, 0, 500))
+	return Context{l: l}
+}
+
 // Logger returns the logger with the context previously set.
 func (c Context) Logger() Logger {
 	return c.l
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/rs/zerolog
+module github.com/madkins23/zerolog
 
 go 1.15
 
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/madkins23/zerolog
+module github.com/rs/zerolog
 
 go 1.15
 


rs:
Would https://github.com/rs/zerolog/pull/575 let you solve your issue?

madkins23:
It fixes part of the problem. The other part would then just require a `NewContext(logger)` function to be made public to create a new `zerolog.Context` and set its private logger field.

rs:
Couldn't you use `UpdateContext` for this?

madkins23:
TL;DR You're right, #575 would solve the more difficult part of my problem and existing methods on `zerolog.Logger` appear to cover the rest.

My [original work-around PR](https://github.com/phsym/zeroslog/pull/4) addressing `v1.32.0` on `phsym/zeroslog` used the `With` method instead. `UpdateContext` is a pointer method, `With` isn't. I was thinking that the `With` method was more likely to return a copy of the original logger as opposed to accidentally modifying it. I admit I didn't do any testing to see if that would happen.

I guess I considered my use of `With` as off-label and possibly dangerous in the long run. Having a `NewContext(logger)` in place seemed more correct and I didn't see #575. But if you're blessing `UpdateContext` (or better yet `With` IMHO) for that usage I'm good with it.

rs:
Yes I think `With` should do the job.

madkins23:
Thanks. Looking forward to #575.

madkins23:
This is no longer necessary AFAIK. It would be nice to get a `v1.33.0` tag to wrap it all up.

rs:
Done

madkins23:
Thanks! I've pushed a PR to `phsym/zeroslog` based on #575. I'm going to close this PR.

rs:
@madkins23 have you considered submitting a PR to integrate your slog support in zerolog itself?

madkins23:
Not really. I'm not actually part of the `phsym/zeroslog` development team so it's not really "my" slog support. I stepped in because I've been using `rs/zerolog` for half a decade and like it very much. I was testing `phsym/zeroslog` when `V1.32.0` was tagged and the `zeroslog` wrapper broke.

I did comment on doing something like this in #571 but [I phrased it as `rs/zerolog` V2 ](https://github.com/rs/zerolog/issues/571#issuecomment-1973760184) which (somewhat predictably in retrospect) didn't go down well. :-(

Let me think about it a bit. I might take a whack at it after I get through with the current set of tweaks to [my test harness](https://github.com/madkins23/go-slog).

PR #662, opened 2024-04-05
Set debug log color

### LINKED ISSUE #661: Give debug level logs a color again
Before v1.32.0 debug logs had a yellow color but that was removed. Vouching to have a color set so they are easily identifiable.

rubenpoppe:
Set default debug level color to magenta

Closes #661

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -128,7 +128,7 @@ var (
 	// log levels.
 	LevelColors = map[Level]int{
 		TraceLevel: colorBlue,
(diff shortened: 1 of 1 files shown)


mitar:
Noo, debug lines should no be so eye catching. That was on purpose that they were less exposed so that it is easier to see those lines which are generally more important.

Anyway, if you are changing this, you should also update the screenshot in README.

PR #663, opened 2024-04-05
Allow setting floating point precision in JSON

### LINKED ISSUE #557: Floating Point Precision
I can understand why  `-1`  was used [here](https://github.com/rs/zerolog/blob/9070d49a1a4154c819e9c5ae6c923f3cc916d959/internal/json/types.go#LL314C44-L314C44), but it can cause some pain when loading the message, say in elasticsearch with the value `0` instead of `0.00` for example.

For me, it creates a field of type `long` instead of what I really want it to be, a `float` and I'm sure there may be other specific cases.

I propose adding a `FloatingPointPrecission` variable somewhere around [here](https://github.com/rs/zerolog/blob/9070d49a1a4154c819e9c5ae6c923f3cc916d959/globals.go#L110) with a default value of `-1`, then everyone is happy :)

mitar:
Fixes #557.

### DIFF
--- a/internal/json/types.go
+++ b/internal/json/types.go
@@ -299,7 +299,7 @@ func (Encoder) AppendUints64(dst []byte, vals []uint64) []byte {
 	return dst
 }
 
-func appendFloat(dst []byte, val float64, bitSize int) []byte {
+func appendFloat(dst []byte, val float64, bitSize, precision int) []byte {
 	// JSON does not permit NaN or Infinity. A typical JSON encoder would fail
 	// with an error, but a logging library wants the data to get through so we
 	// make a tradeoff and store those types as string.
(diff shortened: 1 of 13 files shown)


internal/json/types.go
```diff
@@ -311,16 +311,20 @@ func appendFloat(dst []byte, val float64, bitSize int) []byte {
 	case math.IsInf(val, -1):
 		return append(dst, `"-Inf"`...)
 	}
+	precision := FloatingPointPrecisionFunc()
```
rs, line 314: Would make more sense to pass the precision as an argument.

mitar: So I was not able to set this as a global variable because that would introduce an import loop.

If you are OK with that, I can introduce a parameter to `appendFloat` and then pass it everywhere it is called.

mitar:
@rs: Pushed with the argument.

mitar:
@rs: Done.

PR #665, opened 2024-04-09
Resolve panic in the console writer for short `level` field values

ndk:
This PR fixes an issue in the `zerolog` console writer. Specifying a non-typical value for the `level` field, resulting in a serialized value of less than 3 bytes, causes a panic. For instance:

```golang
log := zerolog.New(zerolog.NewConsoleWriter())
log.Debug().Str("level", "-").Msg("Hello World")
```

### DIFF
--- a/console.go
+++ b/console.go
@@ -28,6 +28,8 @@ const (
 
 	colorBold     = 1
 	colorDarkGray = 90
+
+	unknownLevel = "???"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


console.go
```diff
@@ -401,21 +401,38 @@ func consoleDefaultFormatTimestamp(timeFormat string, location *time.Location, n
 }
 
 func consoleDefaultFormatLevel(noColor bool) Formatter {
+	prefix := func(s string) string {
```
rs, line 404: Why isn't this a top level function?

console.go
```diff
@@ -401,21 +401,38 @@ func consoleDefaultFormatTimestamp(timeFormat string, location *time.Location, n
...
 			fl, ok := FormattedLevels[level]
 			if ok {
 				l = colorize(fl, LevelColors[level], noColor)
 			} else {
-				l = strings.ToUpper(ll)[0:3]
+				if len(ll) == 0 {
+					l = unknownLevel
+				} else {
+					l = strings.ToUpper(prefix(ll))
+				}
```
rs, line 424: What about making all this logic a function instead of just prefix?

ndk: Done :)

console.go
```diff
@@ -401,24 +401,29 @@ func consoleDefaultFormatTimestamp(timeFormat string, location *time.Location, n
 }
 
 func consoleDefaultFormatLevel(noColor bool) Formatter {
+	const unknownLevel = "???"
+	stripLevel := func(ll string) string {
```
rs, line 405: Any reason for this function not to be top level?

ndk: By default, I try to reduce the scope of visibility as much as possible (especially when the stuff is too specific and doesn't look like it's to be reused elsewhere), but I don't have any strong preferences. So, if you think it's better to extract it from the function to the package, I'll do it promptly. :)

rs: I prefer smaller unit functions as they enhance readability and do not eliminate the possibility of inlining.

Issue #493, opened 2022-10-11
[question] How to access `with` fields inside a hook


vijayrajah:
I'm using a hook to write the logs to graylog using GELF client ( "gopkg.in/Graylog2/go-gelf.v2/gelf")

it appears that there is no way to access with fields

like
```
log.With().Str("k","v).Logger().Info().Msg("some message")
```

I want to add the `k` as extra field (with its value as `v`) in GELF message.

is there a way we can do this?

cruzanstx:
I ran into  the same issue. I'm not sure if this is the most performant way to access it and it would be nice if it were built into the event object. I used reflection to access a readonly copy of the buf variable.


```
func (h hook) Run(e *zerolog.Event, level zerolog.Level, msg string){
   logData := make(map[string]interface{})
   // create a string that appends } to the end of the buf variable you access via reflection
   ev := fmt.Sprintf("%s}", reflect.ValueOf(e).Elem().FieldByName("buf"))
   json.Unmarshal([]byte(rv), &logData)

   // now you can either access a map of the data (logData) or string of the data (ev)
}

christopher-taylor:
Because I wound up here from google I would like to add a link for the next person who stumbles across this. Instead of using reflection inside a hook to get at the data, you should consider using an IO.Writer per the example in [this comment](https://github.com/rs/zerolog/pull/682#issuecomment-2395677680).

paologaleotti:
Any updates?

Nucrea:
Same, can't use zerolog together with OpenTelemetry

sruehl:
I use a hook for log event filtering with `.Discard()` and hit the same bump when I was in need to access the `caller` field. So either Event needs to be extended with public read methods or the hook `Run` interface needs to get passed a `ReadableEvent` as the currrent `Event` is write-only.

wgarunap:
I have the same issue, if I use zerolog hook i will lose all my attribute data. Event api needs to be extended on zerolog side. Did anyone had any luck way around it for now?

danielloader:
Can't speak for everyone but I imagine for the OTEL bridging use case the ideal result is access to all attributes on the event, all the way up the potentially multiple `.With()` chaining so you can map them all to attributes on OTEL event submission time.

Issue #587, opened 2023-09-08
How to retrieve logger fields from Zerolog event


jackhua222:
I'm planning to use a hook to send logs to elsewhere on top of the logger. 
I know that the hook interface is like this:
```
type Hook interface {
	// Run runs the hook with the event.
	Run(e *Event, level Level, message string)
}
```
I'm interested to retrieve all the fields I've added to the logger context before sending the logs to elsewhere. I couldn't find anything obvious, is there a way to do this?

doctaweeks:
#493 might help

Issue #637, opened 2024-01-17
Is there a way to obtain the 'buf' from the Event in a hook?


WUST-mengqinyu:
Or alternatively, is it possible to get the 'Context' in the Writer?

RmStorm:
I would like something like this too! I have a hook that adds trace data to the logs. In that (or another) hook I would also like to add bits and pieces of the log event to the trace in cases of high severity. However now it's not possible to lift any information out of the zerolog event. a `.Read()` method on the event would be very nice.

rs:
The buf is an internal structure containing the incomplete event in the making. It is not meant to be exposed.

PR #682, opened 2024-07-22
feat: Add method to get event fields

erezrokah:
Hi 👋 Thank you for this great module 🚀 

Opening for feedback as an attempt to fix https://github.com/rs/zerolog/issues/493, https://github.com/rs/zerolog/issues/587, also referenced from https://github.com/open-telemetry/opentelemetry-go-contrib/pull/5918/files#diff-a9b82897838da6b8eed398e03c9590621e6c2336ff576b3a49fb684ff8bba2a1R110

This is not the cleanest approach, but could be better from using reflecting. It doesn't expose the internal buffer representation format so if it ever changes `GetFields` can be modifed.

Feedback very much welcomed

### DIFF
--- a/event.go
+++ b/event.go
@@ -1,7 +1,9 @@
 package zerolog
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"net"
 	"os"
@@ -827,4 +829,20 @@ func (e *Event) MACAddr(key string, ha net.HardwareAddr) *Event {
 	}
 	e.buf = enc.AppendMACAddr(enc.AppendKey(e.buf, key), ha)
 	return e
+}
+
+func (e *Event) GetFields() (map[string]interface{}, error) {
+	if e == nil {
+		return nil, nil
+	}
+
+	eventFields := make(map[string]interface{})
+	buffer := e.buf
+	if !bytes.HasSuffix(e.buf, []byte("}\n")) {
+		buffer = append(e.buf, '}')
+	}
+	if err := json.Unmarshal(buffer, &eventFields); err != nil {
+		return nil, err
+	}
+	return eventFields, nil
 }
--- a/event_test.go
+++ b/event_test.go
@@ -1,3 +1,4 @@
+//go:build !binary_log
 // +build !binary_log
 
 package zerolog
@@ -61,5 +62,29 @@ func TestEvent_EmbedObjectWithNil(t *testing.T) {
 	got := strings.TrimSpace(buf.String())
 	if got != want {
 		t.Errorf("Event.EmbedObject() = %q, want %q", got, want)
+	}
+}
+
+func TestEvent_GetFields(t *testing.T) {
+	e := newEvent(nil, DebugLevel)
+	e.Str("foo", "bar").Float64("n", 42).Msg("test")
+
+	got, err := e.GetFields()
+	if err != nil {
+		t.Error(err)
+	}
+	want := map[string]interface{}{
+		"foo":     "bar",
+		"n":       float64(42),
+		"message": "test",
+	}
+
+	if len(got) != len(want) {
+		t.Errorf("Event.GetFields() = %v, want %v", len(got), len(want))
+	}
+	for k, v := range want {
+		if got[k] != v {
+			t.Errorf("Event.GetFields() = %v, want %v", got[k], v)
+		}
 	}
 }


event.go
```diff
@@ -828,3 +830,19 @@ func (e *Event) MACAddr(key string, ha net.HardwareAddr) *Event {
...
 }
+
+func (e *Event) GetFields() (map[string]interface{}, error) {
+	if e == nil {
+		return nil, nil
+	}
+
+	eventFields := make(map[string]interface{})
+	buffer := e.buf
+	if !bytes.HasSuffix(e.buf, []byte("}\n")) {
```
ccoVeille, line 841: So if it ends with } but not }\n you will add }\n, so you will have this }}\n and it would be invalid

Maybe you could apply trimspace the buf.

Or check if the prefix is }

ccoVeille, line 841: I don't get the purpose of this. A comment might help

ccoVeille, line 841: This part should be covered by a test

erezrokah:
Thanks for the quick review @ccoVeille, followed up with the suggestions and added more tests. Also please see the comment about the tradeoff of using JSON Decode

event.go
```diff
@@ -828,3 +830,21 @@ func (e *Event) MACAddr(key string, ha net.HardwareAddr) *Event {
...
+// GetFields returns the JSON decoded fields of the event. Please note that the fields might not be mapped to their original type as JSON decoder maps numbers to float64, etc.
+func (e *Event) GetFields() (map[string]interface{}, error) {
+	if e == nil {
+		return nil, nil
+	}
+
+	eventFields := make(map[string]interface{})
+	buffer := bytes.TrimSpace(e.buf)
+	// If `Msg()` was not called the buffer will be missing the closing curly brace
+	if !bytes.HasSuffix(buffer, []byte("}")) {
```
ccoVeille, line 843:
```suggestion
	if !bytes.HasSuffix(buffer, '}')) {
```

ccoVeille, line 843: What about JSON defined by an array

```json
[
{"foo":false}
]
```

The last other options for a valid JSON are

```json
true
```

```json
false
```

```json
null
```

But I'm unsure they could happen in your use case

ccoVeille (commented):
Thanks for applying the changes, here are a few more feedbacks

event.go
```diff
@@ -828,3 +830,21 @@ func (e *Event) MACAddr(key string, ha net.HardwareAddr) *Event {
 	e.buf = enc.AppendMACAddr(enc.AppendKey(e.buf, key), ha)
 	return e
 }
+
+// GetFields returns the JSON decoded fields of the event. Please note that the fields might not be mapped to their original type as JSON decoder maps numbers to float64, etc.
+func (e *Event) GetFields() (map[string]interface{}, error) {
```
ccoVeille, line 835: Is it really about fields if you are getting the values with them?

Maybe GetMetadata or GetJSON would be a better?

erezrokah: `GetMetadata` works I think as we're not getting a JSON (we're parsing a JSON).

event.go
erezrokah: OK tried this and it doesn't seem to work as `HasSuffix` accepts only a `[]byte`

erezrokah: Don't think it can happen https://github.com/rs/zerolog/blob/c78e50e2da70f4ae63e1b65222c3acf12e9ba699/log.go#L490 https://github.com/rs/zerolog/blob/c78e50e2da70f4ae63e1b65222c3acf12e9ba699/internal/json/types.go#L18

but that's why I'm returning an error as the parsing can fail for other reasons as well

event_test.go
```diff
@@ -63,3 +64,63 @@ func TestEvent_EmbedObjectWithNil(t *testing.T) {
 		t.Errorf("Event.EmbedObject() = %q, want %q", got, want)
 	}
 }
+
+func TestEvent_GetFields(t *testing.T) {
```
ccoVeille, line 68: Don't forget to rename this

ccoVeille (approved):

event_test.go
erezrokah: Done in https://github.com/rs/zerolog/pull/682/commits/f6d674dbec4b018409abf57e38092f1f2b05c83e

ccoVeille (approved):
LGTM 🤞

chkp-omris:
@ccoVeille @erezrokah I'd also be very happy to have this feature merged, I happened to find this PR looking for a solution.
May I suggest that the implementation simply returns the internal buffer `buf` to avoid forcing JSON decoding on the user?

erezrokah:
> @ccoVeille @erezrokah I'd also be very happy to have this feature merged, I happened to find this PR looking for a solution.
> May I suggest that the implementation simply returns the internal buffer `buf` to avoid forcing JSON decoding on the user?

I following https://github.com/rs/zerolog/issues/637#issuecomment-2036756423 as not to expose the internal `buf`. If we expose `buf` and the later there's a need to change the internal representation of the event to something else it will be harder to do

mark-ten9:
It would be great to merge this if possible so that we don't have to use reflection to work around this.

nduong-ol:
Hello, it has been a while. Would it be possible to merge this PR?

christopher-taylor:
Chiming in to say I would _love_ this feature to be merged in!

nduong-ol:
In the end I find that to implement io.Writer interface receive the log output from zerolog and export it with OTLP is quite effective also. We can get the intended formatted log from zerolog without the need to process the internal data ourself

christopher-taylor:
> In the end I find that to implement io.Writer interface receive the log output from zerolog and export it with OTLP is quite effective also. We can get the intended formatted log from zerolog without the need to process the internal data ourself
Do you have a minimal example implementation?

nduong-ol:
> > In the end I find that to implement io.Writer interface receive the log output from zerolog and export it with OTLP is quite effective also. We can get the intended formatted log from zerolog without the need to process the internal data ourself
> 
> Do you have a minimal example implementation?

Sure. Here the code of the logger with the same approach of the Hook but instead of receiving zerolog.Event, it takes the byte slice from logger output and export it.
`package logging

import (
	"context"
	"io"

	"go.opentelemetry.io/otel/log"
	"go.opentelemetry.io/otel/log/global"
)

type config struct {
	provider  log.LoggerProvider
	version   string
	schemaURL string
}

func newConfig(options []Option) config {
	var c config
	for _, opt := range options {
		c = opt.apply(c)
	}

	if c.provider == nil {
		c.provider = global.GetLoggerProvider()
	}
	return c
}

func (c config) logger(name string) log.Logger {
	var opts []log.LoggerOption
	if c.version != "" {
		opts = append(opts, log.WithInstrumentationVersion(c.version))
	}
	if c.schemaURL != "" {
		opts = append(opts, log.WithSchemaURL(c.schemaURL))
	}
	return c.provider.Logger(name, opts...)
}

// Option configures a Hook.
type Option interface {
	apply(config) config
}
type optFunc func(config) config

func (f optFunc) apply(c config) config { return f(c) }

// WithVersion returns an [Option] that configures the version of the
// [log.Logger] used by a [OtelLogger]. The version should be the version of the
// package that is being logged.
func WithVersion(version string) Option {
	return optFunc(func(c config) config {
		c.version = version
		return c
	})
}

// WithSchemaURL returns an [Option] that configures the semantic convention
// schema URL of the [log.Logger] used by a [OtelLogger]. The schemaURL should be
// the schema URL for the semantic conventions used in log records.
func WithSchemaURL(schemaURL string) Option {
	return optFunc(func(c config) config {
		c.schemaURL = schemaURL
		return c
	})
}

// WithLoggerProvider returns an [Option] that configures [log.LoggerProvider]
//
// By default if this Option is not provided, the Logger will use the global LoggerProvider.
func WithLoggerProvider(provider log.LoggerProvider) Option {
	return optFunc(func(c config) config {
		c.provider = provider
		return c
	})
}

var _ io.WriteCloser = (*OtelLogger)(nil)

type OtelLogger struct {
	logger log.Logger
}

func NewOtelLogger(name string, options ...Option) *OtelLogger {
	cfg := newConfig(options)
	return &OtelLogger{
		logger: cfg.logger(name),
	}
}

func (l *OtelLogger) Write(p []byte) (n int, err error) {
	r := log.Record{}
	r.SetSeverity(log.SeverityInfo)
	r.SetBody(log.StringValue(string(p)))
	r.SetSeverityText("INFO")

	l.logger.Emit(context.Background(), r)
	return len(p), nil
}

// Close implements io.Closer, and closes the current logfile.
func (l *OtelLogger) Close() error {
	return nil
}
`
 And the integration with zerolog
 `	var writers []io.Writer {
		zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: time.RFC3339},
                NewOtelLogger("app-name"),
	}

	logLevel := zerolog.InfoLevel
	logContext := zerolog.New(io.MultiWriter(writers...)).Level(logLevel).With().Timestamp()
`

christopher-taylor:
> > > In the end I find that to implement io.Writer interface receive the log output from zerolog and export it with OTLP is quite effective also. We can get the intended formatted log from zerolog without the need to process the internal data ourself
> > 
> > 
> > Do you have a minimal example implementation?
> 
> Sure. Here the code of the logger with the same approach of the Hook but instead of receiving zerolog.Event, it takes the byte slice from logger output and export it. The approach is similar on how we would write zerolog to a file. In this approach I didn't unmarshal the json log to get the log level because I'm lazy. If you want to can unmarshal the zerolog json to extract information before sending it over OTLP
> 
> ```go
> package` logging
> 
> import (
> 	"context"
> 	"io"
> 
> 	"go.opentelemetry.io/otel/log"
> 	"go.opentelemetry.io/otel/log/global"
> )
> 
> type config struct {
> 	provider  log.LoggerProvider
> 	version   string
> 	schemaURL string
> }
> 
> func newConfig(options []Option) config {
> 	var c config
> 	for _, opt := range options {
> 		c = opt.apply(c)
> 	}
> 
> 	if c.provider == nil {
> 		c.provider = global.GetLoggerProvider()
> 	}
> 	return c
> }
> 
> func (c config) logger(name string) log.Logger {
> 	var opts []log.LoggerOption
> 	if c.version != "" {
> 		opts = append(opts, log.WithInstrumentationVersion(c.version))
> 	}
> 	if c.schemaURL != "" {
> 		opts = append(opts, log.WithSchemaURL(c.schemaURL))
> 	}
> 	return c.provider.Logger(name, opts...)
> }
> 
> // Option configures a Hook.
> type Option interface {
> 	apply(config) config
> }
> type optFunc func(config) config
> 
> func (f optFunc) apply(c config) config { return f(c) }
> 
> // WithVersion returns an [Option] that configures the version of the
> // [log.Logger] used by a [OtelLogger]. The version should be the version of the
> // package that is being logged.
> func WithVersion(version string) Option {
> 	return optFunc(func(c config) config {
> 		c.version = version
> 		return c
> 	})
> }
> 
> // WithSchemaURL returns an [Option] that configures the semantic convention
> // schema URL of the [log.Logger] used by a [OtelLogger]. The schemaURL should be
> // the schema URL for the semantic conventions used in log records.
> func WithSchemaURL(schemaURL string) Option {
> 	return optFunc(func(c config) config {
> 		c.schemaURL = schemaURL
> 		return c
> 	})
> }
> 
> // WithLoggerProvider returns an [Option] that configures [log.LoggerProvider]
> //
> // By default if this Option is not provided, the Logger will use the global LoggerProvider.
> func WithLoggerProvider(provider log.LoggerProvider) Option {
> 	return optFunc(func(c config) config {
> 		c.provider = provider
> 		return c
> 	})
> }
> 
> var _ io.WriteCloser = (*OtelLogger)(nil)
> 
> type OtelLogger struct {
> 	logger log.Logger
> }
> 
> func NewOtelLogger(name string, options ...Option) *OtelLogger {
> 	cfg := newConfig(options)
> 	return &OtelLogger{
> 		logger: cfg.logger(name),
> 	}
> }
> 
> func (l *OtelLogger) Write(p []byte) (n int, err error) {
> 	r := log.Record{}
> 	r.SetSeverity(log.SeverityInfo)
> 	r.SetBody(log.StringValue(string(p)))
> 	r.SetSeverityText("INFO")
> 
> 	l.logger.Emit(context.Background(), r)
> 	return len(p), nil
> }
> 
> // Close implements io.Closer, and closes the current logfile.
> func (l *OtelLogger) Close() error {
> 	return nil
> }
> ```
> 
> And the integration with zerolog
> 
> ```
>        var writers []io.Writer {
> 		zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: time.RFC3339},
>                 NewOtelLogger("app-name"),
> 	}
> 
> 	logLevel := zerolog.InfoLevel
> 	logContext := zerolog.New(io.MultiWriter(writers...)).Level(logLevel).With().Timestamp()
> ```

Thanks! I have linked this to another issue I arrived at from google so future searchers arrive here instead of the answer that involves costly reflection.

PeterBocan:
@erezrokah @ccoVeille what's the status of this PR?

ekristen:
hey @rs apologies for the direct ping however I didn't see a better choice. It's been well over a year since this was opened, and not a lot of activity on the repo in the last 5 months.

Love this library, it's definitely my preferred, would like to see this get merged, so I wanted to call this to your attention and also ask, is there a better way or a preferred way to get maintainer involvement on PRs like this?

Thanks again for the library, much appreciated.

nicholasheyer:
> > Can we get eyes on this @IDisposable
> 
> I'm honestly waiting on a bunch of my own PRs to get merged, and for the ones that have already been merged to be packaged in a release. I suspect @rs has just been overwhelmed with _real life_. I will help in any way I can, of course, but I'm just a contributor, no write-access. Sorry.

No worries and apologies for the random tag. I may be looking for alternative logging libraries :(

PR #702, opened 2025-01-01
Update LICENSE, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛


rs:
The license keep the original date as long as no major change happened. No need to update it every year or a range should be used.

PR #705, opened 2025-01-16
Props field in event to handle log properties in hooks

SaravananSai07:


rs:
This change would allocate and kill all the benefits of the way this lib is designed.

SaravananSai07:
> This change would allocate and kill all the benefits of the way this lib is designed.

Figured. Making an alternative.

PR #706, opened 2025-01-17
Exposed get values on Event to access logged param while handling hooks

SaravananSai07:


rs:
This method won't work until the event is written as the JSON in the buffer won't be completed/valid. What is the use-case of this?

SaravananSai07:
> ~This method won't work until the event is written as the JSON in the buffer won't be completed/valid.~ What is the use-case of this?

Selectively publish logs to another destination based on keys passed in the log. And this happens in a hook.

rs:
This approach is highly inefficient and involves unnecessary allocations. Exposing such a function could encourage its use and create the misleading impression that it is an efficient solution. In my opinion, using a logging library as an event system is not the right design choice. A more effective approach would be to introduce a dedicated layer between your emitters and the logging library. This layer can handle event dispatching to other systems as needed, ensuring a cleaner and more efficient separation of concerns.

SaravananSai07:
> This approach is highly inefficient and involves unnecessary allocations. Exposing such a function could encourage its use and create the misleading impression that it is an efficient solution. In my opinion, using a logging library as an event system is not the right design choice. A more effective approach would be to introduce a dedicated layer between your emitters and the logging library. This layer can handle event dispatching to other systems as needed, ensuring a cleaner and more efficient separation of concerns.

Allocations are an issue but it is made sure that it happens only if someone needs the props. Not to be defensive, if the library provide hooks and let the clients act based on level and message, why not other params in the log?

rs:
Ideally, you could route or filter messages based on any field, not just levels. However, due to the performance-oriented design of this library, this is not possible without parsing the buffer you just built. I’d say that if you need to do that, you’d be better off with a library that isn’t designed with zero allocation and performance as priorities. Such a library would likely end up being more optimized for your use case than parsing the buffer.

PR #713, opened 2025-03-02
Expose the internal function cbor.DecodeIfBinaryToBytes()

marshallflax:
cbor.DecodeIfBinaryToBytes() is a useful function, but it's inside zerolog/internal and therefore not accessible directly. And it's only implicitly available when the `binary_log` build tag is enabled. I'd like to be able to use it in downstream log collectors, but currently the only approach which works is to cut-and-paste the code into a local module.  Adding a publically-visible wrapper would make this easy.

Thanks!

### DIFF
--- a/encoder.go
+++ b/encoder.go
@@ -55,7 +55,6 @@ type encoder interface {
 	AppendUints8(dst []byte, vals []uint8) []byte
 }
 
-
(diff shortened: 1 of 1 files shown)


rs:
Why don't you use a cbor library? This is internal to avoid a dependency. It wouldn't make sense to expose it as it is not the purpose of this library.

PR #716, opened 2025-03-14
(PR) Added `.HexBytes` method

SilvioRispoliAtElectroluxDotCom:
It acts pretty much like `.Hex`, but instead of building a single string containing the hex repr of the bytes, it returns the array repr of hex bytes (eg: [0x12, 0x34, ...]).

Note: I've sneaked in the upcase version of the hex runes, probably can merge it with the internal strings and use a config switch in order to choose (up/down)case output. So I'm certain this will not be approved from the get-go, but please consider adding such feature to mainstream, adapting it as needed


rs:
I’m failing to think about a use case for this. Can you help me understand?

SilvioRispoliAtElectroluxDotCom:
Sure. It's just another way of printing a sequence of bytes.

I cannot delve too much into the details of what I'm logging.. Think of frames traveling on a communication bus.
Something that can look like this: `[<header>, <id>, <data...>, <checksum>, <footer>]`
Or with some fake data: `[0xFA, 0x10, 0x12, 0x34, 0x56, 0x60, 0xAF]`

One could use the already existing `.Hex` method to add such frames to the log, but I had two issues:
- the hex runes are printed lowercase, with no option for upcasing them
- the frame would be printed as such: `fa1012345660af` which might have some application elsewhere, but I wanted to have them split like in the example above

So, it's mere preference, can totally understand why you can't see yourself using it 😄

rs:
I'm trying to avoid bloating the API of this lib with various ways of formatting the same type of data. The format proposed would inflate a lot the log events, which is likely not something most people would like. You can build those arrays manually, without the lib needing to implement this format.

SilvioRispoliAtElectroluxDotCom:
> I'm trying to avoid bloating the API of this lib with various ways of formatting the same type of data. 

It might be the same type of data, but it is not the same kind. Plus, I like options, and making this change was the only way that Go can tolerate.

> The format proposed would inflate a lot the log events, which is likely not something most people would like. 

.. And that would be my problem, not yours 😄. Any of the available formats can inflate the log if used not properly, but that's the user's fault, not yours nor `zerolog`'s fault

> You can build those arrays manually, without the lib needing to implement this format.

But that's the whole point. I need that exact format

rs:
My point is that you can create this format without changing the library

rs:
Using `Strs` and a function on your side to en/decode a byte[] to this format.

Issue #397, opened 2022-01-16
Add ability to override exit function


nickpalmer:
There should be a way to override the os.Exit() function call with a different exit handler.

```
zerolog.SetExitFunc(exit func(int) {})
// or even via a global
var ExitFunc = os.Exit()
```

This would allow people to do required cleanup such as closing a log file, or to write tests that call `Fatal()` and check that fatal was called.

roger-russel:
That would be nice because it would allow me to test fatal functions.

kieronlanning:
Was there ever a solution for this? Looking into how I can write unit tests around a function that uses the `log.Fatal()` function.

IDisposable:
Coming in PR #748

PR #717, opened 2025-03-20
Add ExitFunc for FatalLevel log messages

FriedEgg:
resolves rs/zerolog#397

inspiration from logrus...

This change is primarily motivated by the need to facilitate easier testing in applications that frequently use `Fatal()`. Additionally, it could benefit scenarios where applications require custom shutdown procedures for cleanup or other special handling but still choose to rely on zerolog to perform process termination.

### DIFF
--- a/globals.go
+++ b/globals.go
@@ -3,7 +3,6 @@ package zerolog
 import (
 	"bytes"
 	"encoding/json"
-	"os"
 	"strconv"
 	"sync/atomic"
 	"time"
@@ -121,9 +120,6 @@ var (
 	// be thread safe and non-blocking.
 	ErrorHandler func(err error)
 
-	// ExitFunc is the fucntion called to exit the application, defaults to `os.Exit()`
-	ExitFunc exitFunc = os.Exit
-
 	// DefaultContextLogger is returned from Ctx() if there is no logger associated
 	// with the context.
 	DefaultContextLogger *Logger
@@ -166,8 +162,6 @@ var (
 	gLevel          = new(int32)
 	disableSampling = new(int32)
 )
-
-type exitFunc func(int)
 
 // SetGlobalLevel sets the global override for log level. If this
 // values is raised, all Loggers will use at least this value.
--- a/globals.go
+++ b/globals.go
@@ -3,6 +3,7 @@ package zerolog
 import (
 	"bytes"
 	"encoding/json"
+	"os"
 	"strconv"
 	"sync/atomic"
 	"time"
@@ -120,6 +121,9 @@ var (
 	// be thread safe and non-blocking.
 	ErrorHandler func(err error)
 
+	// ExitFunc is the fucntion called to exit the application, defaults to `os.Exit()`
+	ExitFunc exitFunc = os.Exit
+
 	// DefaultContextLogger is returned from Ctx() if there is no logger associated
 	// with the context.
 	DefaultContextLogger *Logger
@@ -162,6 +166,8 @@ var (
 	gLevel          = new(int32)
 	disableSampling = new(int32)
 )
+
+type exitFunc func(int)
 
 // SetGlobalLevel sets the global override for log level. If this
 // values is raised, all Loggers will use at least this value.
--- a/log.go
+++ b/log.go
@@ -221,19 +221,25 @@ func (l Level) MarshalText() ([]byte, error) {
 	return []byte(LevelFieldMarshalFunc(l)), nil
 }
 
+// ExitFunc is a function type that takes an integer exit code and terminates the program.
+// By default, it is set to os.Exit, but it can be overridden to customize the behavior
+// of program termination (for example, for graceful shutdown or testing purposes).
+type ExitFunc func(int)
+
 // A Logger represents an active logging object that generates lines
 // of JSON output to an io.Writer. Each logging operation makes a single
 // call to the Writer's Write method. There is no guarantee on access
 // serialization to the Writer. If your Writer is not thread safe,
 // you may consider a sync wrapper.
 type Logger struct {
-	w       LevelWriter
-	level   Level
-	sampler Sampler
-	context []byte
-	hooks   []Hook
-	stack   bool
-	ctx     context.Context
+	w        LevelWriter
+	level    Level
+	sampler  Sampler
+	context  []byte
+	hooks    []Hook
+	stack    bool
+	ctx      context.Context
+	exitFunc ExitFunc
 }
 
 // New creates a root logger with given output writer. If the output writer implements
@@ -335,6 +341,12 @@ func (l Logger) Hook(hooks ...Hook) Logger {
 	return l
 }
 
+// ExitFunc returns a logger with the e ExitFunc.
+func (l Logger) ExitFunc(e ExitFunc) Logger {
+	l.exitFunc = e
+	return l
+}
+
 // Trace starts a new message with trace level.
 //
 // You must call Msg on the returned event in order to send the event.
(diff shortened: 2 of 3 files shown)


globals.go
```diff
@@ -120,6 +121,9 @@ var (
 	// be thread safe and non-blocking.
 	ErrorHandler func(err error)
 
+	// ExitFunc is the fucntion called to exit the application, defaults to `os.Exit()`
```
ccoVeille, line 124: 🖊️  typo

```suggestion
	// ExitFunc is the function called to exit the application, defaults to `os.Exit()`
```

globals.go
```diff
@@ -120,6 +121,9 @@ var (
...
+	ExitFunc exitFunc = os.Exit
```
ccoVeille, line 125: You are not using `exitFunc` except here and you didn't add a comment.

So I don't think you use this type, so I would like to suggest this

```suggestion
	ExitFunc func(int) = os.Exit
```

So I'm also suggesting to delete the exitFunc type [here](https://github.com/rs/zerolog/pull/717#discussion_r2006417117)

globals.go
```diff
@@ -163,6 +167,8 @@ var (
 	disableSampling = new(int32)
 )
 
+type exitFunc func(int)
```
ccoVeille, line 170: Please see here why I suggest to remove this

https://github.com/rs/zerolog/pull/717#discussion_r2006416883

```suggestion
```

Also having an non-exported type on ExitFunc that is q globally exported func is a bad thing

globals.go
ccoVeille, line 125: I see a real problem here, with the implementation

You are doing this change for test purpose, but you are introducing a global exiter that anyone can overload.

It shouldn't be global and exported.

This would lead to unpredictable behavior with tests, but also with code.

Think about someone anywhere in log that would do a logger.ExitFunc = func(_ int) {}

log.go
```diff
@@ -382,18 +382,22 @@ func (l *Logger) Err(err error) *Event {
...
+			// will be lost as ExitFunc() (os.Exit by default) terminates the program immediately.
 			closer.Close()
 		}
-		os.Exit(1)
+		if ExitFunc == nil {
+			ExitFunc = os.Exit
+		}
+		ExitFunc(1)
 	})
 }
```
ccoVeille, line 402: About the problem I raised [here](https://github.com/rs/zerolog/pull/717#discussion_r2006423308), here is what I suggest you to do

```go
type ExitFunc func(int)

type Logger struct {
        w       LevelWriter
        level   Level
        sampler Sampler
        context []byte
        hooks   []Hook
        stack   bool
        ctx     context.Context
        exiter ExitFunc
}
```

```suggestion
		exitFunc = l.exitFunc
		if exitFunc == nil {
			exitFunc = os.Exit
		}
		exitFunc(1)
 	})
 }

func(l *logger) SetTestExiter(t *testing.T, exitFunc ExitFunc) {
	l.exitFunc = exitFunc
	t.Cleanup(func() {
		l.exitFunc = nil
	})
}
```

This is a safe way, there won't be data race, no global overlapping between packages calling module

ccoVeille (changes requested):
While I understand the need, I'm afraid the implementation could be improved

Please accept my feedbacks and review

log.go
FriedEgg: super valid criticisms, thanks for the feedback

FriedEgg: for the setter, it seems more inline with other functions to have a function such as the following, to return a new child Logger instance with the desired ExitFunc than to modify it?

```go
// ExitFunc returns a logger with the e ExitFunc.
func (l Logger) ExitFunc(e ExitFunc) Logger {
	l.exiter = e
	return l
}
```

the intention wasn't to make this functionality exclusively available for testing purposes so I am not sure I see a reason to add the suggest `SetTestExiter`? Would having both make sense in your eyes?

FriedEgg: I have pushed some modifications based on your feedback. For the time being I have chosen to not tie the setting of `l.exitFunc` to testing and will wait for additional feedback from others.

log_test.go
```diff
@@ -78,6 +81,57 @@ func TestInfo(t *testing.T) {
...
+		}
+	})
+
+	t.Run("should not exit", func(t *testing.T) {
+		out := &bytes.Buffer{}
+		log := New(out).ExitFunc(func(_ int) {})
+		log.Fatal().Msg("")
+		if got, want := decodeIfBinaryToString(out.Bytes()), `{"level":"fatal"}`+"\n"; got != want {
+			t.Errorf("invalid log output:\n got:  %v\nwant: %v", got, want)
+		}
```
ccoVeille, line 122: Adding such a thing would help to validate the exiter is called

```suggestion
		out := &bytes.Buffer{}
		var receivedCode int
		log := New(out).ExitFunc(func(code int) {
			receivedCode = code
		})
		log.Fatal().Msg("")
		if got, want := receivedCode, 1; got != want {
			t.Errorf("invalid exitCode:\n got:  %v\nwant: %v", got, want)
		}
		if got, want := decodeIfBinaryToString(out.Bytes()), `{"level":"fatal"}`+"\n"; got != want {
			t.Errorf("invalid log output:\n got:  %v\nwant: %v", got, want)
		}
```

log_test.go
```diff
@@ -78,6 +81,57 @@ func TestInfo(t *testing.T) {
...
+	})
+
+	t.Run("with level fatal should not exit", func(t *testing.T) {
+		out := &bytes.Buffer{}
+		log := New(out)
+		log.WithLevel(FatalLevel).Msg("")
+		if got, want := decodeIfBinaryToString(out.Bytes()), `{"level":"fatal"}`+"\n"; got != want {
+			t.Errorf("invalid log output:\n got:  %v\nwant: %v", got, want)
+		}
+	})
```
ccoVeille, line 132: You are not setting the exiter

I'm unsure what you are testing here

log_test.go
```diff
@@ -78,6 +81,57 @@ func TestInfo(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		out, err := io.ReadAll(stderr)
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = cmd.Wait()
+		if err == nil {
+			t.Error("Expected log Fatal to exit with non-zero status")
+		}
```
ccoVeille, line 109: You should take a look at

- https://pkg.go.dev/os/exec#Cmd.Wait

- https://pkg.go.dev/os/exec#ExitError

- https://pkg.go.dev/os#ProcessState.ExitCode


This way you could to something like this


```suggestion
		err = cmd.Wait()
		if err == nil {
			t.Error("Expected test to fail")
		}

		var errExit exec.ExitError
		if !errors.As(err, &errExit) {
			t.Error("Expected an os.ExitError")
		}

		if errExit.ExitCode() != 1 {
			t.Error("Expected log Fatal to exit with non-zero status")
		}
```

ccoVeille:
Also nitpicking but a Co-authored-by tag would be more appreciated than a simple mention in the commit message.

rs:
Thanks for the proposal.

A Fatal should not be used in most cases and certainly not in testable code.

Adding a func pointer in the logger is not something I’m willing to add for this as it adds more weight to a pass by copy type.

Sorry to disappoint but I’m not convinced this is a necessary addition.

FriedEgg:
No disappointment here it can certainly be lived without especially if only for the benefit of testing.

ccoVeille:
I accept the argument of memory consumption. It would add extra memory for anyone using the lib. For a feature that would be rarely used.

@FriedEgg good PR and discussion anyway.

PR #732, opened 2025-10-03
fix: Context.EmbedObject returns a broken json when object is empty

rerorero:
This PR fixes an issue with Context.EmbedObject() that produces invalid JSON when
- the given object's `MarshalZerologobject` adds no fields (empty object)
- and the parent context has existing fields.

When `MarshalZerologObject` adds no fields (e.g., all fields are conditionally skipped), an extra comma is inserted:

```json
  {"level":"info","parent_field":"parent_value",,"message":"test"}
```


IDisposable (approved):
Looks good :)

rs:
Dup with #737?

IDisposable:
Yes, I did have the same thing fixed as this in #737. You can merge this one first and I'll update... or just merge #737 as it covers this case as well.

PR #733, opened 2025-10-23
feat: add new global DurationFieldFormat

### LINKED ISSUE #709: Console formatting for duration
Console output is meant to be human-readable, but durations are shown as floats (e.g., in seconds or milliseconds) which is far from human-readable for non-simple durations. I know that the issue is that console output is interpreting data in JSON, and in JSON the fact that a field was a duration is lost, is just a number, but I still think some solution to this should be provided.

One way would be to use `FormatFieldValue` but sadly it does not get field name as a parameter, so one cannot hard-code it to (known/used) duration field names to convert float to formatted duration string.

Maybe `DurationFields` could get a list of duration field names?

Currently, it looks that the only workaround is to use `FormatPrepare` which can do some preprocessing based on (known/used) field names.

Any other suggestion how to tackle this?

sruehl:
This deprecates DurationFieldInteger in favor of DurationFieldFormat with the value of DurationFormatInt

resolves #709


sruehl:
@rs did a short local test. Could not find a good place to put a unit tests for it but seems to work just fine.

globals.go
```diff
@@ -107,12 +117,16 @@ var (
...
+	// DurationFieldFormat defines the format of the Duration field type.
+	DurationFieldFormat = DurationFormatFloat
+
 	// DurationFieldUnit defines the unit for time.Duration type fields added
 	// using the Dur method.
 	DurationFieldUnit = time.Millisecond
 
 	// DurationFieldInteger renders Dur fields as integer instead of float if
 	// set to true.
+	// Deprecated: use DurationFieldFormat with DurationFormatInt instead.
```
ccoVeille, line 129:
```suggestion
	// Deprecated: use [DurationFieldFormat] with [DurationFormatInt] instead.
```

PR #734, opened 2025-12-05
Add array handling of IPAddr and IPPrefix

IDisposable:
- Updated documentation, benchmark, and tests 
- Caught up some missing documentation and benchmarks.
- Cleaned up couple lint messages

PR #735, opened 2025-12-06
Chore improve test coverage

IDisposable:
While putting in the unit tests for #734 I realized that there was some gaps in the unit test coverage... so I added a bunch.

- Added Type to the BenchmarkLogFieldType and BenchmarkContexdtFieldType test cases, also to log_test.go
- Moved the test-fixture data for them and also shared them with new event test for nil events.
- Added a couple strings for encodeByteTest for FF, CR, LF
- Added complete cbor testing for AppendInt*, AppendInts*, AppendUint*, and AppendUints* methods.
- Added test for float IsNaN and Inf(0) and Inf(-1).
- Added all min/max/smallest value tests for ints and floats
- Extended the Array test to include IPPrefix, MACAddr, Interface, and object marshaled.
- Better test of MarshalZerologObject.
- Added test for encoder AppendStrings and AppendStringers.

** THIS BUILDS OVER #734** ~Don't merge this if you don't want that PR. Or merge~ Since you merged that one ~first and then I'll make~ I am sure there are no merge issues.

### DIFF
--- a/internal/json/float_test.go
+++ b/internal/json/float_test.go
@@ -1,0 +1,433 @@
+package json
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"math"
+	"math/rand"
+	"testing"
+
+	"github.com/rs/zerolog/internal"
+)
+
+var float64Tests = []struct {
+	Name string
+	Val  float64
+	Want string
+}{
+	{
+		Name: "Positive integer",
+		Val:  1234.0,
+		Want: "1234",
+	},
+	{
+		Name: "Negative integer",
+		Val:  -5678.0,
+		Want: "-5678",
+	},
+	{
+		Name: "Positive decimal",
+		Val:  12.3456,
+		Want: "12.3456",
+	},
+	{
+		Name: "Negative decimal",
+		Val:  -78.9012,
+		Want: "-78.9012",
+	},
+	{
+		Name: "Large positive number",
+		Val:  123456789.0,
+		Want: "123456789",
+	},
+	{
+		Name: "Large negative number",
+		Val:  -987654321.0,
+		Want: "-987654321",
+	},
+	{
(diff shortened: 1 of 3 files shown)


internal/cbor/types_test.go
```diff
@@ -100,21 +108,217 @@ var integerTestCases = []struct {
...
+	if !bytes.Equal(got, want) {
+		t.Errorf("AppendInts8(%v)\ngot:  0x%s\nwant: 0x%s",
+			array, hex.EncodeToString(got),
+			hex.EncodeToString(want))
+	}
+
+	// now empty array case
+	array = make([]int8, 0)
+	want = make([]byte, 0)
+	want = append(want, 0x9f) // start and end array
```
IDisposable: All empty arrays emit these two bytes.

internal/cbor/types_test.go
```diff
@@ -128,12 +332,308 @@ func TestAppendInt(t *testing.T) {
...
+			hex.EncodeToString(want))
+	}
+}
+
+func TestAppendUint(t *testing.T) {
+	for _, tc := range unsignedIntegerTestCases {
+		s := enc.AppendUint([]byte{}, tc.val)
+		got := string(s)
+		want := tc.bigbinary
+		if tc.val == math.MaxUint64 {
```
IDisposable: The cool optimization that assigned a single-byte representation to Max-Unsigned-Native-Int

benchmark_test.go
```diff
@@ -115,115 +101,88 @@ func BenchmarkLogArrayObject(b *testing.B) {
 }
 
 func BenchmarkLogFieldType(b *testing.B) {
-	bools := []bool{true, false, true, false, true, false, true, false, true, false}
```
IDisposable: Moved this data to fixtures_test.go to be shared

PR #737, opened 2025-12-17
Fix missing context.Context in Object/EmbedObject

### LINKED ISSUE #730: EmbedObject doesn't set the context.Context on the intermediate event is uses to apply MarshalZerologObject with
This line from https://github.com/rs/zerolog/blob/9dacc014f38d60f563c2ab18719aec11fc06765c/context.go#L69

```
// EmbedObject marshals and Embeds an object that implement the LogObjectMarshaler interface.
func (c Context) EmbedObject(obj LogObjectMarshaler) Context {
	e := newEvent(LevelWriterAdapter{io.Discard}, 0)
	e.EmbedObject(obj)
	c.l.context = enc.AppendObjectData(c.l.context, e.buf)
	putEvent(e)
	return c
}
```

Creates a newEvent and then runs EmbedObject immediately before returning the Context. The problem is that it does not pass c.l.ctx to the new event, which means that LogObjectMarshalers that do something like this,

```
func (embed *embedContextKeys) MarshalZerologObject(event *zerolog.Event) {
	context := event.GetCtx()
        ...
}
```
Will have always have a nil context if called from `EmbedObject()` vs being called from `Event.EmbedObject(..)`

I can't tell if this is the intended design, but it seems odd since the ctx could be set via `Ctx()` before `EmbedObject()` is called.

IDisposable:
Fixes #730
Copies over event settings for ctx, hooks, and stack
~Does **NOT** address Array.Object or Event.Dict.~

- Fix missing context for `Event.appendObject`, passing the event's stack, ctx, and hooks (ch).
- Add `Array.Errs` support.
- `Context.Errs`, `Event.Errs`,  use new `Array.Errs()`
- `Array.Err` handles nil correctly.
- Fix `Event.Err` code that ignored the modified event in `ErrorStackMarshaler` handling.
- Fix `Fields.appendFieldList` when `ErrorMarshalFunc` returns a nil.
- `Fields.appendFieldList` no longer adds the `ErrorStackFieldName` field when `ErrorStackMarshaler` returns `nil`.
- Don't call the `ErrorMarshalFunc` on `nil` `error`s.
- Removed unreachable code NOP in `Context.Array`
- Add `Context.CreateArray()` and `Event.CreateArray()` to deprecate `Array.Arr()` func so contextual data is carried into `Array` objects
- Add `Context.CreateDict()` and `Event.CreateDict()` to deprecate `Event.Dict()` func so contextual data is carried into the `Dict()`'s `Event` for `appendObject`'s use
- Rewrite `ErrorMarshalFunc` testing.

### DIFF
--- a/array.go
+++ b/array.go
@@ -1,6 +1,7 @@
 package zerolog
 
 import (
+	"context"
 	"net"
 	"sync"
 	"time"
@@ -17,7 +18,10 @@ var arrayPool = &sync.Pool{
 // Array is used to prepopulate an array of items
 // which can be re-used to add to log messages.
 type Array struct {
-	buf []byte
+	buf   []byte
+	stack bool            // enable error stack trace
+	ctx   context.Context // Optional Go context
+	ch    []Hook          // hooks
 }
 
 func putArray(a *Array) {
@@ -31,13 +35,22 @@ func putArray(a *Array) {
 	if cap(a.buf) > maxSize {
 		return
 	}
+	a.stack = false
+	a.ctx = nil
+	a.ch = nil
 	arrayPool.Put(a)
 }
 
 // Arr creates an array to be added to an Event or Context.
+// WARNING: This function is deprecated because it does not preserve
+// the stack, hooks, and context from the parent event.
+// Deprecated: Use Event.CreateArray or Context.CreateArray instead.
 func Arr() *Array {
 	a := arrayPool.Get().(*Array)
 	a.buf = a.buf[:0]
+	a.stack = false
+	a.ctx = nil
+	a.ch = nil
 	return a
 }
 
@@ -59,8 +72,7 @@ func (a *Array) write(dst []byte) []byte {
 // Object marshals an object that implement the LogObjectMarshaler
 // interface and appends it to the array.
 func (a *Array) Object(obj LogObjectMarshaler) *Array {
-	// TODO can we get context, stack, and hooks here?
-	a.buf = appendObject(enc.AppendArrayDelim(a.buf), obj, false, nil, nil)
+	a.buf = appendObject(enc.AppendArrayDelim(a.buf), obj, a.stack, a.ctx, a.ch)
 	return a
 }
 
@@ -109,24 +121,24 @@ func (a *Array) Err(err error) *Array {
 }
 
 // Errs serializes and appends errors to the array.
-func (arr *Array) Errs(errs []error) *Array {
+func (a *Array) Errs(errs []error) *Array {
 	for _, err := range errs {
 		switch m := ErrorMarshalFunc(err).(type) {
 		case nil:
-			arr = arr.Interface(nil)
+			a = a.Interface(nil)
 		case LogObjectMarshaler:
-			arr = arr.Object(m)
+			a = a.Object(m)
 		case error:
 			if !isNilValue(m) {
-				arr = arr.Str(m.Error())
+				a = a.Str(m.Error())
 			}
 		case string:
-			arr = arr.Str(m)
+			a = a.Str(m)
 		default:
-			arr = arr.Interface(m)
+			a = a.Interface(m)
 		}
 	}
-	return arr
+	return a
 }
 
 // Bool appends the val as a bool to the array.
--- a/context.go
+++ b/context.go
@@ -34,8 +34,28 @@ func (c Context) Dict(key string, dict *Event) Context {
 	return c
 }
 
+// CreateDict creates an Event to be used with the Context.Dict method.
+// It preserves the stack, hooks, and context from the logger.
+// Call usual field methods like Str, Int etc to add fields to this
+// event and give it as argument the Context.Dict method.
+func (c Context) CreateDict() *Event {
+	return newEvent(nil, DebugLevel, c.l.stack, c.l.ctx, c.l.hooks)
+}
+
+// CreateArray creates an Array to be used with the Context.Array method.
+// It preserves the stack, hooks, and context from the logger.
+// Call usual field methods like Str, Int etc to add elements to this
+// array and give it as argument the Context.Array method.
+func (c Context) CreateArray() *Array {
+	a := Arr()
+	a.stack = c.l.stack
+	a.ctx = c.l.ctx
+	a.ch = c.l.hooks
+	return a
+}
+
 // Array adds the field key with an array to the event context.
-// Use zerolog.Arr() to create the array or pass a type that
+// Use c.CreateArray() to create the array or pass a type that
 // implement the LogArrayMarshaler interface.
 func (c Context) Array(key string, arr LogArrayMarshaler) Context {
 	c.l.context = enc.AppendKey(c.l.context, key)
@@ -43,7 +63,7 @@ func (c Context) Array(key string, arr LogArrayMarshaler) Context {
 		c.l.context = arr.write(c.l.context)
 		return c
 	}
-	a := Arr()
+	a := c.CreateArray()
 	arr.MarshalZerologArray(a)
 	c.l.context = a.write(c.l.context)
 	return c
@@ -155,7 +175,7 @@ func (c Context) AnErr(key string, err error) Context {
 // Errs adds the field key with errs as an array of serialized errors to the
 // logger context.
 func (c Context) Errs(key string, errs []error) Context {
-	arr := Arr().Errs(errs)
+	arr := c.CreateArray().Errs(errs)
 	return c.Array(key, arr)
 }
 
--- a/error_marshal_test.go
+++ b/error_marshal_test.go
@@ -35,6 +35,10 @@ func (w wrappedError) Error() string {
 	return w.error.Error() + ": " + w.msg
 }
 
+type interfaceError struct {
+	val string
+}
+
 func TestArrayErrorMarshalFunc(t *testing.T) {
 	prefixed := func(s, prefix string) string {
 		if s == "null" {
@@ -90,7 +94,19 @@ func TestArrayErrorMarshalFunc(t *testing.T) {
 			want: []string{`null`, `"failure"`, `"whoops"`, `"oops"`},
 		},
 		{
-			name: "nil error",
+			name: "interface",
+			marshal: func(err error) interface{} {
+				var some interfaceError
+				if err != nil {
+					some.val = err.Error()
+				}
+				var interfaceErr interface{} = some
+				return interfaceErr
+			},
+			want: []string{`{}`, `{}`, `{}`, `{}`},
+		},
+		{
+			name: "nilError",
 			marshal: func(err error) interface{} {
 				var errNil error = nil
 				return errNil
@@ -130,7 +146,7 @@ func TestArrayErrorMarshalFunc(t *testing.T) {
 						wants := `[` + want + `]`
 						a := Arr().Err(err)
 						if got := decodeObjectToStr(a.write([]byte{})); got != wants {
-							t.Errorf("%s Array.Err(%v)\ngot:  %s\nwant: %s", tc.name, err, got, wants)
+							t.Errorf("%s %d Array.Err(%v)\ngot:  %s\nwant: %s", tc.name, i, err, got, wants)
 						}
 					})
 					t.Run("Ctx", func(t *testing.T) {
@@ -139,7 +155,7 @@ func TestArrayErrorMarshalFunc(t *testing.T) {
 						logger := New(out).With().Err(err).Logger()
 						logger.Log().Msg("msg")
 						if got := decodeIfBinaryToString(out.Bytes()); got != wants {
-							t.Errorf("%s Ctx.Err(%v)\ngot:  %v\nwant: %v", tc.name, err, got, wants)
+							t.Errorf("%s %d Ctx.Err(%v)\ngot:  %v\nwant: %v", tc.name, i, err, got, wants)
 						}
 					})
 					t.Run("Event", func(t *testing.T) {
@@ -148,16 +164,19 @@ func TestArrayErrorMarshalFunc(t *testing.T) {
 						logger := New(out)
 						logger.Log().Err(err).Msg("msg")
 						if got := decodeIfBinaryToString(out.Bytes()); got != wants {
-							t.Errorf("%s Event.Err(%v)\ngot:  %v\nwant: %v", tc.name, err, got, wants)
+							t.Errorf("%s %d Event.Err(%v)\ngot:  %v\nwant: %v", tc.name, i, err, got, wants)
 						}
 					})
 					t.Run("Fields", func(t *testing.T) {
+						if i == 0 && tc.want[i] == "{}" {
+							want = `null`
+						}
 						wants := `{"err":` + want + `,"message":"msg"}` + "\n"
 						out := &bytes.Buffer{}
 						logger := New(out)
 						logger.Log().Fields(map[string]interface{}{"err": err}).Msg("msg")
 						if got := decodeIfBinaryToString(out.Bytes()); got != wants {
-							t.Errorf("%s Event.Fields(%v)\ngot:  %v\nwant: %v", tc.name, err, got, wants)
+							t.Errorf("%s %d Event.Fields(%v)\ngot:  %v\nwant: %v", tc.name, i, err, got, wants)
 						}
 					})
 				}
--- a/event.go
+++ b/event.go
@@ -170,7 +170,7 @@ func (e *Event) Fields(fields interface{}) *Event {
 }
 
 // Dict adds the field key with a dict to the event context.
-// Use zerolog.Dict() to create the dictionary.
+// Use e.CreateDict() to create the dictionary.
 func (e *Event) Dict(key string, dict *Event) *Event {
 	if e == nil {
 		return e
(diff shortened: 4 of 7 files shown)


array.go
```diff
@@ -59,7 +59,8 @@ func (a *Array) write(dst []byte) []byte {
 // Object marshals an object that implement the LogObjectMarshaler
 // interface and appends it to the array.
 func (a *Array) Object(obj LogObjectMarshaler) *Array {
-	a.buf = appendObject(enc.AppendArrayDelim(a.buf), obj)
+	// TODO can we get context, stack, and hooks here?
```
IDisposable: @rs not sure how (or if it's reasonable) to get this information for Array.Object

event.go
```diff
@@ -184,7 +185,8 @@ func (e *Event) Dict(key string, dict *Event) *Event {
 // Call usual field methods like Str, Int etc to add fields to this
 // event and give it as argument the *Event.Dict method.
 func Dict() *Event {
-	return newEvent(nil, 0)
+	// TODO can we get context, stack, and hooks here?
```
IDisposable: @rs not sure how (or if it's reasonable) to get this information for Event.Dict as it's not called against an Event instance (so there's nowhere to copy the contextual  information from...)

log.go
```diff
@@ -487,22 +487,21 @@ func (l *Logger) newEvent(level Level, done func(string)) *Event {
...
 		e.buf = enc.AppendObjectData(e.buf, l.context)
 	}
-	if l.stack {
-		e.Stack()
-	}
 	return e
 }
 
+func (l *Logger) scratchEvent() *Event {
+	return newEvent(LevelWriterAdapter{io.Discard}, DebugLevel, l.stack, l.ctx, l.hooks)
```
IDisposable, line 502: ...just shorthand so context doesn't have to _know_ about what all logger-level attributes to copy over

array.go
```diff
@@ -90,12 +91,12 @@ func (a *Array) RawJSON(val []byte) *Array {
 // Err serializes and appends the err to the array.
 func (a *Array) Err(err error) *Array {
 	switch m := ErrorMarshalFunc(err).(type) {
+	case nil:
```
IDisposable, line 106: Coverage tests prove that this will be called if the ErrorMarshalFunc returns a `nil` (untyped) or `error`-typed `nil` so handle those here.

array.go
```diff
@@ -90,12 +91,12 @@ func (a *Array) RawJSON(val []byte) *Array {
...
+		a.buf = enc.AppendNil(enc.AppendArrayDelim(a.buf))
 	case LogObjectMarshaler:
 		a = a.Object(m)
 	case error:
-		if m == nil || isNilValue(m) {
-			a.buf = enc.AppendNil(enc.AppendArrayDelim(a.buf))
-		} else {
+		if !isNilValue(m) {
```
IDisposable, line 111: The `case nil:` will catch actual `nil`s, this only catches things that seem to be an error, but are actually zero-object.

array.go
```diff
@@ -107,6 +108,27 @@ func (a *Array) Err(err error) *Array {
 	return a
 }
 
+// Errs serializes and appends errors to the array.
```
IDisposable, line 123: Centralized all the handling of Errs here, since everyone was calling `Arr()` and then doing the same things inline.

array_test.go
```diff
@@ -47,14 +47,14 @@ func TestArray(t *testing.T) {
 		).
 		Err(nil).
 		Err(fmt.Errorf("failure")).
-		Err(errorObjectMarshalerImpl{fmt.Errorf("oops")}).
+		Err(loggableError{fmt.Errorf("oops")}).
```
IDisposable, line 50: Unified the test classes for derived `error`s

context.go
```diff
@@ -23,7 +23,7 @@ func (c Context) Logger() Logger {
 // Only map[string]interface{} and []interface{} are accepted. []interface{} must
 // alternate string keys and arbitrary values, and extraneous ones are ignored.
 func (c Context) Fields(fields interface{}) Context {
-	c.l.context = appendFields(c.l.context, fields, c.l.stack)
+	c.l.context = appendFields(c.l.context, fields, c.l.stack, c.l.ctx, c.l.hooks)
```
IDisposable, line 25: Pass the rest of the context that might affect the `LogObjectMarshal` call, specifically added the `context.Context` in that is in the `zerolog.Context`'s logger. Same with the hooks stored in the `*zerolog.Logger`.

context.go
```diff
@@ -44,20 +44,15 @@ func (c Context) Array(key string, arr LogArrayMarshaler) Context {
 		c.l.context = arr.write(c.l.context)
 		return c
 	}
-	var a *Array
-	if aa, ok := arr.(*Array); ok {
```
IDisposable, line 48: This code is unreachable, and also did nothing. The test is the same as the one [line 43](https://github.com/rs/zerolog/pull/737/changes#diff-552f47512a00afe5fc6850cc9ddc830a6daeca162750e50aab4ed549685e0253R43) so it would never be true. That means we're actually always going to execute `a = Arr()` and fall down to the write.

context.go
```diff
@@ -44,20 +44,15 @@ func (c Context) Array(key string, arr LogArrayMarshaler) Context {
...
+	a := Arr()
+	arr.MarshalZerologArray(a)
 	c.l.context = a.write(c.l.context)
 	return c
 }
 
 // Object marshals an object that implement the LogObjectMarshaler interface.
 func (c Context) Object(key string, obj LogObjectMarshaler) Context {
-	e := newEvent(LevelWriterAdapter{io.Discard}, 0)
+	e := c.l.scratchEvent()
```
IDisposable, line 74: Centralized the creation of "scratch" `Event` and moved to the `Logger`. 

We need to make sure we set the `Event`'s copies of the `context.Context` (in `.ctx`, ), stack-dump flag (in `.stack`), and the hooks (in `.hooks` for `Logger`, in `.ch` for `Event`)

All instances use `io.Discard` and 0 == `zerolog.DebugLevel`.

context.go
```diff
@@ -161,24 +156,7 @@ func (c Context) AnErr(key string, err error) Context {
...
-				arr = arr.Str(m.Error())
-			}
-		case string:
-			arr = arr.Str(m)
-		default:
-			arr = arr.Interface(m)
-		}
-	}
-
+	arr := Arr().Errs(errs)
```
IDisposable: Use code in `Array.Errs()`

ctx_test.go
```diff
@@ -82,14 +82,6 @@ func (t logObjectMarshalerImpl) MarshalZerologObject(e *Event) {
 	e.Str("name", strings.ToLower(t.name)).Int("age", -t.age)
 }
 
-type errorObjectMarshalerImpl struct {
```
IDisposable, line 85: Moved to `loggableError` from the (new)  _error_marshal_test.go_ file

error_marshal_test.go
```diff
@@ -0,0 +1,205 @@
+package zerolog
```
IDisposable, line 1: Complete testing of all things ErrorMarshalFunc related!

event.go
```diff
@@ -57,14 +57,15 @@ type LogArrayMarshaler interface {
 	MarshalZerologArray(a *Array)
 }
 
-func newEvent(w LevelWriter, level Level) *Event {
+func newEvent(w LevelWriter, level Level, stack bool, ctx context.Context, hooks []Hook) *Event {
```
IDisposable, line 60: This is the main thing changing here... we copy the logging context information that could be important to the `LogObjectMarshaler`... mostly we care about the `context.Context` in `.ctx`, but really should be copying the hooks and stack flag as well.

event.go
```diff
@@ -360,11 +362,10 @@ func (e *Event) AnErr(key string, err error) *Event {
 	case LogObjectMarshaler:
 		return e.Object(key, m)
 	case error:
-		if m == nil || isNilValue(m) {
+		if isNilValue(m) {
```
IDisposable, line 392: `case nil:` would have caught an actual `nil`

event.go
```diff
@@ -378,20 +379,7 @@ func (e *Event) Errs(key string, errs []error) *Event {
...
-		case error:
-			arr = arr.Err(m)
-		case string:
-			arr = arr.Str(m)
-		default:
-			arr = arr.Interface(m)
-		}
-	}
-
+	arr := Arr().Errs(errs)
```
IDisposable: Leverage code from `Array.Errs`

event.go
```diff
@@ -407,21 +395,22 @@ func (e *Event) Err(err error) *Event {
...
 		return e
 	}
+
 	if e.stack && ErrorStackMarshaler != nil {
 		switch m := ErrorStackMarshaler(err).(type) {
 		case nil:
+			// do nothing
 		case LogObjectMarshaler:
-			e.Object(ErrorStackFieldName, m)
+			e = e.Object(ErrorStackFieldName, m)
```
IDisposable, line 431: Fix the lost `*Event` returned from these methods, when the old code continued at line 414, it was running against the original `*Event`... technically, this doesn't break anything because event is actually modifying the context's buffers, but this is the only place I noticed where we're not following the fluent conventions... could bite if there is ever contextual/state information added to the `Event`

event_test.go
```diff
@@ -50,30 +56,104 @@ func TestEvent_writeWithNil(t *testing.T) {
...
+		}
+		val := ctx.Value(contextKey)
+		if val == nil {
+			t.Errorf("expected context value, got %v", val)
+		}
+		e.Str("ctxValue", val.(string))
+		e.Bool("stackValue", e.stack)
+	})
+
+	t.Run("ObjectWithFullContext", func(t *testing.T) {
```
IDisposable, line 118: Added tests for `context.Context` access, for example getting a value in a hook function that was stored in the context via `WithValue`

fields.go
```diff
@@ -28,46 +30,42 @@ func appendFields(dst []byte, fields interface{}, stack bool) []byte {
...
+			dst = appendFieldList(dst, kv, stack, ctx, hooks)
 		}
 	}
 	return dst
 }
 
-func appendObject(dst []byte, obj LogObjectMarshaler) []byte {
-	e := newEvent(nil, 0)
+func appendObject(dst []byte, obj LogObjectMarshaler, stack bool, ctx context.Context, hooks []Hook) []byte {
+	e := newEvent(LevelWriterAdapter{io.Discard}, DebugLevel, stack, ctx, hooks)
```
IDisposable, line 40: Can't use `Logger`'s `scratchEvent` here because this is not bound to an `*Event`.

fields.go
```diff
@@ -77,29 +75,33 @@ func appendFieldList(dst []byte, kvList []interface{}, stack bool) []byte {
 			}
 
 			if stack && ErrorStackMarshaler != nil {
-				dst = enc.AppendKey(dst, ErrorStackFieldName)
 				switch m := ErrorStackMarshaler(val).(type) {
 				case nil:
+					// do nothing
+				case LogObjectMarshaler:
```
IDisposable, line 81: This instance didn't have handling for when a `LogObjectMarshaler` was returned from `ErrorStackMarshaler` like all the other places.

fields.go
```diff
@@ -77,29 +75,33 @@ func appendFieldList(dst []byte, kvList []interface{}, stack bool) []byte {
 			}
 
 			if stack && ErrorStackMarshaler != nil {
-				dst = enc.AppendKey(dst, ErrorStackFieldName)
```
IDisposable, line 80: We no longer add a key when the marshaled error stack is `nil`. In the old code, this was a bare key, which seems invalid (the null was not being added).

fields.go
```diff
@@ -28,46 +30,42 @@ func appendFields(dst []byte, fields interface{}, stack bool) []byte {
...
-func appendFieldList(dst []byte, kvList []interface{}, stack bool) []byte {
+func appendFieldList(dst []byte, kvList []interface{}, stack bool, ctx context.Context, hooks []Hook) []byte {
 	for i, n := 0, len(kvList); i < n; i += 2 {
 		key, val := kvList[i], kvList[i+1]
 		if key, ok := key.(string); ok {
 			dst = enc.AppendKey(dst, key)
 		} else {
 			continue
 		}
-		if val, ok := val.(LogObjectMarshaler); ok {
```
IDisposable, line 54: This early code messes with handling of `LogObjectMarshaler` processed fields. Most of the time, they were added as `{}` and none of the embedded properties. Now handled below in the `default:` case

fields.go
```diff
@@ -294,7 +296,11 @@ func appendFieldList(dst []byte, kvList []interface{}, stack bool) []byte {
 		case json.RawMessage:
 			dst = appendJSON(dst, val)
 		default:
-			dst = enc.AppendInterface(dst, val)
+			if lom, ok := val.(LogObjectMarshaler); ok {
```
IDisposable, line 299: This handles the case from the deleted code above... as previously coded it would short-circuit other types that "just happen" to satisfy the interface. Now we do it only if the type _itself_ isn't handled by other cases.

log.go
```diff
@@ -487,22 +487,21 @@ func (l *Logger) newEvent(level Level, done func(string)) *Event {
 		}
 		return nil
 	}
-	e := newEvent(l.w, level)
+	e := newEvent(l.w, level, l.stack, l.ctx, l.hooks)
```
IDisposable, line 490: Passing on the context

log_test.go
```diff
@@ -163,7 +163,7 @@ func TestWithPlurals(t *testing.T) {
 		Strs("strings_nil", nil).
 		Stringers("stringers", []fmt.Stringer{net.IP{127, 0, 0, 1}, nil}).
 		Stringers("stringers_nil", nil).
-		Errs("errs", []error{errors.New("some error"), errors.New("some other error"), nil, errorObjectMarshalerImpl{fmt.Errorf("oops")}}).
+		Errs("errs", []error{errors.New("some error"), errors.New("some other error"), nil, loggableError{fmt.Errorf("oops")}, nonLoggableError{fmt.Errorf("whoops"), 401}}).
```
IDisposable, line 166: More options for better coverage.

log_test.go
```diff
@@ -863,363 +864,6 @@ func TestOutputWithTimestamp(t *testing.T) {
 	}
 }
 
-type loggableError struct {
```
IDisposable, line 866: Moved to _error_marshal_test.go_

IDisposable:
@rs I think this is ready to go now... fixed the random sync-related issues in the _diode_test.go_ which were randomly breaking the  test builds.

fixtures_test.go
```diff
@@ -120,7 +120,7 @@ func makeFieldFixtures() *fieldFixtures {
 	ipPfxV6 := net.IPNet{IP: net.IP{0x20, 0x01, 0x0d, 0xb8, 0x85, 0xa3, 0x00, 0x00, 0x00, 0x00, 0x8a, 0x2e, 0x03, 0x70, 0x73, 0x00}, Mask: net.CIDRMask(64, 128)}
 	ipPfxs := []net.IPNet{ipPfxV4, ipPfxV6, ipPfxV4, ipPfxV6, ipPfxV4, ipPfxV6, ipPfxV4, ipPfxV6, ipPfxV4, ipPfxV6}
 	macAddr := net.HardwareAddr{0x00, 0x1A, 0x2B, 0x3C, 0x4D, 0x5E}
-	errs := []error{errors.New("a"), errors.New("b"), errors.New("c"), errors.New("d"), errors.New("e"), nil, errorObjectMarshalerImpl{fmt.Errorf("oops")}}
+	errs := []error{errors.New("a"), errors.New("b"), errors.New("c"), errors.New("d"), errors.New("e"), nil, loggableError{fmt.Errorf("oops")}}
```
IDisposable, line 123: More coverage.

diode/diode_test.go
```diff
@@ -60,15 +61,23 @@ func TestFatal(t *testing.T) {
...
-	if err != nil {
-		t.Fatal(err)
-	}
+
+	var stderrBuf bytes.Buffer
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		_, _ = io.Copy(&stderrBuf, stderr)
```
IDisposable: Turns out that to reliably capture the output, we have to copy it out asynchronously (and before the command ends), and we need to wait for the read to complete.

array.go
```diff
@@ -107,6 +108,27 @@ func (a *Array) Err(err error) *Array {
...
+func (arr *Array) Errs(errs []error) *Array {
+	for _, err := range errs {
+		switch m := ErrorMarshalFunc(err).(type) {
+		case nil:
+			arr = arr.Interface(nil)
+		case LogObjectMarshaler:
+			arr = arr.Object(m)
+		case error:
+			if !isNilValue(m) {
+				arr = arr.Str(m.Error())
```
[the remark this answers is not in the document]
IDisposable: What say you @rs? This is what it did already, but CoPilot has a point... in either case, it's a moot point because I can seem to get code-coverage on that clause no matter what I do :(

event.go
```diff
@@ -184,7 +185,8 @@ func (e *Event) Dict(key string, dict *Event) *Event {
 // Call usual field methods like Str, Int etc to add fields to this
 // event and give it as argument the *Event.Dict method.
 func Dict() *Event {
-	return newEvent(nil, 0)
+	// TODO can we get context, stack, and hooks here?
+	return newEvent(nil, DebugLevel, false, nil, nil)
```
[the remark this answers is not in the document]
IDisposable: Yeppers... @rs I need some viewpoint on what we should do here... do we want to figure out where to get contextual information or just document there is none? I can do either but would love your perspective here

array.go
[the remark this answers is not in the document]
IDisposable: Same as above... the comment looks nice if we do decide to not determine where we could get context from...

array.go
```diff
@@ -90,12 +91,12 @@ func (a *Array) RawJSON(val []byte) *Array {
 // Err serializes and appends the err to the array.
 func (a *Array) Err(err error) *Array {
 	switch m := ErrorMarshalFunc(err).(type) {
+	case nil:
+		a.buf = enc.AppendNil(enc.AppendArrayDelim(a.buf))
 	case LogObjectMarshaler:
 		a = a.Object(m)
 	case error:
-		if m == nil || isNilValue(m) {
-			a.buf = enc.AppendNil(enc.AppendArrayDelim(a.buf))
```
IDisposable, line 97: Nothing I could do in testing would ever hit this line. It simply doesn't ever not match the `case nil:`

array.go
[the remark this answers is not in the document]
rs: Can't you repro with something like that? https://go.dev/play/p/ejJvRAC-XBp

event.go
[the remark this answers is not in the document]
rs: I don't see how we could without changing the API.

event.go
rs: You can have a nil type matching the error interface. See my other message.

array.go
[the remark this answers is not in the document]
IDisposable: Will do that this evening

event.go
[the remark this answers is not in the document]
IDisposable: The only thing I could conceive is an ArrayWithContext that you could opt into...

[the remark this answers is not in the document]
IDisposable: The only thing I could conceive is an ArrayWithContext that you could use when you cared...but I really don't know how much value that adds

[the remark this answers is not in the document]
IDisposable: The only thing I could conceive is an ArrayWithContext that you could use when you cared...but I really don't know how much value that adds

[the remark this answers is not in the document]
rs: Or add something like a CreateArray and CreateDict method to Event that populates all the fields and deprecate the other constructors.

array.go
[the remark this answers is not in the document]
IDisposable: I've added tests for interface{} and now have 84% coverage.

array.go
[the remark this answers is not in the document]
IDisposable: Implemented CreateArray and CreateDict against Context and Event so we can always have context. Dict() and Arr() are marked as deprecated.

array.go
IDisposable: > Nothing I could do in testing would ever hit this line. It simply doesn't ever not match the `case nil:`

Now have **all** of these tested _except_ for the `Context` version. Still can't seem to hit that one, but `Event`, `Fields`, and `Array` are now covered.

array.go
IDisposable: Per suggestion, added `Context.CreateArray()` and `Event.CreateArray()` and deprecate `Arr()`

PR #739, opened 2025-12-17
ConsoleWriter can now to emit fields in original order

### LINKED ISSUE #727: ConsoleWriter mangles JSON order
ConsoleWriter reorders JSON fields when printing them, even if the field was added via RawJSON().

It should instead preserve the order of those fields.

IDisposable:
Fixes #727
Set `DoNotSortFields` to only move the fields named in `FieldsOrder` to the front and leave the rest in order added.
Defaults to false

### DIFF
--- a/console.go
+++ b/console.go
@@ -76,6 +76,9 @@ type ConsoleWriter struct {
 	// FieldsOrder defines the order of contextual fields in output.
 	FieldsOrder []string
 
+	// DoNotSortFields does not sort contextual fields alphabetically (default false).
+	DoNotSortFields bool
+
 	fieldIsOrdered map[string]int
 
 	// FieldsExclude defines contextual fields to not display in output.
@@ -101,9 +104,9 @@ type ConsoleWriter struct {
 // NewConsoleWriter creates and initializes a new ConsoleWriter.
 func NewConsoleWriter(options ...func(w *ConsoleWriter)) ConsoleWriter {
 	w := ConsoleWriter{
-		Out:          os.Stdout,
-		TimeFormat:   consoleDefaultTimeFormat,
-		PartsOrder:   consoleDefaultPartsOrder(),
+		Out:        os.Stdout,
+		TimeFormat: consoleDefaultTimeFormat,
+		PartsOrder: consoleDefaultPartsOrder(),
 	}
 
 	for _, opt := range options {
@@ -204,33 +207,9 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
 		fields = append(fields, field)
 	}
 
-	if len(w.FieldsOrder) > 0 {
-		w.orderFields(fields)
-	} else {
-		sort.Strings(fields)
-	}
+	w.orderFields(fields)
 
-	// Write space only if something has already been written to the buffer, and if there are fields.
-	if buf.Len() > 0 && len(fields) > 0 {
-		buf.WriteByte(' ')
-	}
-
-	// Move the "error" field to the front
-	ei := sort.Search(len(fields), func(i int) bool { return fields[i] >= ErrorFieldName })
-	if ei < len(fields) && fields[ei] == ErrorFieldName {
-		fields[ei] = ""
-		fields = append([]string{ErrorFieldName}, fields...)
-		var xfields = make([]string, 0, len(fields))
-		for _, field := range fields {
-			if field == "" { // Skip empty fields
-				continue
-			}
-			xfields = append(xfields, field)
-		}
-		fields = xfields
-	}
-
-	for i, field := range fields {
+	for _, field := range fields {
 		var fn Formatter
 		var fv Formatter
 
@@ -260,6 +239,10 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
 			}
 		}
 
+		// Write space only if something has already been written to the buffer
+		if buf.Len() > 0 {
+			buf.WriteByte(' ')
+		}
 		buf.WriteString(fn(field))
 
 		switch fValue := evt[field].(type) {
@@ -278,10 +261,6 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
 			} else {
 				fmt.Fprint(buf, fv(b))
 			}
-		}
-
-		if i < len(fields)-1 { // Skip space for last field
-			buf.WriteByte(' ')
 		}
 	}
 }
@@ -351,7 +330,7 @@ func (w ConsoleWriter) writePart(buf *bytes.Buffer, evt map[string]interface{}, 
 
 // orderFields takes an array of field names and an array representing field order
 // and returns an array with any ordered fields at the beginning, in order,
-// and the remaining fields after in their original order.
+// and the remaining fields after in their original order or sorted alphabetically.
 func (w ConsoleWriter) orderFields(fields []string) {
 	if w.fieldIsOrdered == nil {
 		w.fieldIsOrdered = make(map[string]int)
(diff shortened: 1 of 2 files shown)


console.go
```diff
@@ -204,33 +207,9 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
...
 	}
 
-	if len(w.FieldsOrder) > 0 {
-		w.orderFields(fields)
-	} else {
-		sort.Strings(fields)
-	}
-
-	// Write space only if something has already been written to the buffer, and if there are fields.
-	if buf.Len() > 0 && len(fields) > 0 {
```
IDisposable, line 214: This got moved to inside the loop before emitting the field

console.go
```diff
@@ -260,6 +239,10 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
 			}
 		}
 
+		// Write space only if something has already been written to the buffer
```
IDisposable, line 242: Moved to inside the loop to eliminate the duplication of before loop and end if loop below.

console.go
```diff
@@ -279,10 +262,6 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
 				fmt.Fprint(buf, fv(b))
 			}
 		}
-
```
IDisposable, line 282: Now done before writing the field.

console.go
```diff
@@ -351,17 +330,43 @@ func (w ConsoleWriter) writePart(buf *bytes.Buffer, evt map[string]interface{},
...
 		for i, fieldName := range w.FieldsOrder {
 			w.fieldIsOrdered[fieldName] = i
 		}
 	}
-	sort.Slice(fields, func(i, j int) bool {
-		ii, iOrdered := w.fieldIsOrdered[fields[i]]
-		jj, jOrdered := w.fieldIsOrdered[fields[j]]
+
+	indexes := w.fieldIsOrdered
+	tiebreaker := func(i, j int) bool {
```
IDisposable, line 343: This is the original tie-breaker on the name of the field.

console.go
```diff
@@ -351,17 +330,43 @@ func (w ConsoleWriter) writePart(buf *bytes.Buffer, evt map[string]interface{},
...
+		}
+
+		for i, field := range fields {
+			if _, ok := indexes[field]; !ok {
+				indexes[field] = offset + i
+			}
+		}
+
+		tiebreaker = func(i, j int) bool {
+			panic("unreachable")
```
IDisposable, line 361: Since all fields actually in the message are in the map, the tiebreaker is never going to get invoked.

console.go
```diff
@@ -351,17 +330,43 @@ func (w ConsoleWriter) writePart(buf *bytes.Buffer, evt map[string]interface{},
...
+		}
+	}
+
+	indexes[ErrorFieldName] = -1 // Always put error field first
+
+	sort.SliceStable(fields, func(i, j int) bool {
```
IDisposable, line 367: Switch to `SliceStable` to ensure we don't rearrange things that don't need to move at all.

console.go
```diff
@@ -351,17 +330,43 @@ func (w ConsoleWriter) writePart(buf *bytes.Buffer, evt map[string]interface{},
...
+
+	indexes := w.fieldIsOrdered
+	tiebreaker := func(i, j int) bool {
+		return fields[i] < fields[j]
+	}
+
+	if w.DoNotSortFields {
+		offset := len(w.fieldIsOrdered)
+		indexes = make(map[string]int, offset)
+		for fieldName, index := range w.fieldIsOrdered {
```
IDisposable, line 350: Makes a copy of the user-specified FieldOrder as we're going to add all the other fields.

console.go
```diff
@@ -351,17 +330,43 @@ func (w ConsoleWriter) writePart(buf *bytes.Buffer, evt map[string]interface{},
...
+			indexes[fieldName] = index
+		}
+
+		for i, field := range fields {
+			if _, ok := indexes[field]; !ok {
+				indexes[field] = offset + i
```
IDisposable, line 356: Place them _at least_ after all the ordered fields

console.go
```diff
@@ -204,33 +207,9 @@ func (w ConsoleWriter) writeFields(evt map[string]interface{}, buf *bytes.Buffer
...
-	}
-
-	// Write space only if something has already been written to the buffer, and if there are fields.
-	if buf.Len() > 0 && len(fields) > 0 {
-		buf.WriteByte(' ')
-	}
+	w.orderFields(fields)
 
-	// Move the "error" field to the front
-	ei := sort.Search(len(fields), func(i int) bool { return fields[i] >= ErrorFieldName })
```
IDisposable, line 219: Eliminate the need for this adjustment by always calling `orderFields`

console.go
```diff
@@ -351,17 +330,43 @@ func (w ConsoleWriter) writePart(buf *bytes.Buffer, evt map[string]interface{},
...
+				indexes[field] = offset + i
+			}
+		}
+
+		tiebreaker = func(i, j int) bool {
+			panic("unreachable")
+		}
+	}
+
+	indexes[ErrorFieldName] = -1 // Always put error field first
```
IDisposable, line 365: Force the field matching `ErrorFieldName` to the front. This eliminates the entire rearrangement block in `writeFields`

console_test.go
```diff
@@ -533,20 +546,37 @@ func TestConsoleWriterConfiguration(t *testing.T) {
...
 
-		evt := `{"level": "info", "message": "Zoo", "aardvark": "Able", "mussel": "Mountain", "zebra": "Zulu"}`
+		evt := `{"level": "info", "message": "Zoo", "aardvark": "Able", "mussel": "Mountain", "zebra": "Zulu", "lizard": "Larry"}`
 		_, err := w.Write([]byte(evt))
 		if err != nil {
 			t.Errorf("Unexpected error when writing output: %s", err)
 		}
 
-		expectedOutput := "<nil> INF Zoo zebra=Zulu aardvark=Able mussel=Mountain\n"
+		expectedOutput := "<nil> INF Zoo zebra=Zulu aardvark=Able lizard=Larry mussel=Mountain\n"
```
IDisposable, line 559: Ensure there are multiple extra fields so we can see them get sorted.

console_test.go
```diff
@@ -533,20 +546,37 @@ func TestConsoleWriterConfiguration(t *testing.T) {
...
 		actualOutput := buf.String()
 		if actualOutput != expectedOutput {
-			t.Errorf("Unexpected output %q, want: %q", actualOutput, expectedOutput)
+			t.Errorf("Unexpected output\ngot:  %q\nwant: %q\n", actualOutput, expectedOutput)
+		}
+	})
+
+	t.Run("Sets FieldsOrder unsorted extras", func(t *testing.T) {
+		buf := &bytes.Buffer{}
+		w := zerolog.ConsoleWriter{Out: buf, NoColor: true, FieldsOrder: []string{"zebra", "aardvark"}, DoNotSortFields: true}
```
IDisposable, line 568: Passing `DoNotSortFields: true` keeps fields in original order _except as far as `FieldsOrder` mentions_.

console_test.go
```diff
@@ -649,6 +679,6 @@ func BenchmarkConsoleWriter(b *testing.B) {
 	w := zerolog.ConsoleWriter{Out: io.Discard, NoColor: false}
 
 	for i := 0; i < b.N; i++ {
-		w.Write(msg)
+		_, _ = w.Write(msg)
```
IDisposable, line 682: _nit_ just got rid of the warning

IDisposable:
Well, after much agony with the random failures of the unit tests, it turns out that we cannot guarantee the order of the fields actually matches the input order of the underlying Event. Even in unit tests where we explicitly give a JSON to parse into an event, the order of the field can change because the implementation in Write does a json.Decode into a map[string]interface{} which means that we don't have field ordering information of the fields. This is because enumerating a map[string] does NOT come out in any particular order...depending on how the map tree was constructed, which can be influenced by the whole-runtime memory state :(. It's clear that json ordering is not a guarantee as RFC 8259 defines an object as an "unordered collection". Meanwhile [RFC 8785 declares that canonical JSON have their object properties sorted](https://www.rfc-editor.org/rfc/rfc8785.html#section-3.2.3).

We could modify the code in console.go to pass around the deserialized object (from event bytes) as an interface{} (e.g. any), but that would change the signature of ConsoleWriter plug-ins FormatPrepare and FormatExtra... which would be a breaking change.

NOTE: The dependency on json.Decode will result in additional issues when and if the upcoming [json v2 enhancements](https://go.dev/blog/jsonv2-exp) come into default use (currently available in go1.25 under the GOEXPERIMENT=jsonv2 build flag.

The new parser is more strict:

It will detect and give an error with duplicate keys (which [zerolog explicitly allows](https://github.com/rs/zerolog?tab=readme-ov-file#field-duplication)).
It will reject invalid UTF-8 which might have been silently ignored before, though you can optionally [allow invalid UTF-8](https://pkg.go.dev/encoding/json/jsontext#AllowInvalidUTF8)
In fact, the v2 will explicitly support [ordered json objects](https://pkg.go.dev/encoding/json/v2#example-package-OrderedObject) and allow [duplicate names](https://pkg.go.dev/encoding/json/jsontext#AllowDuplicateNames).

IDisposable:
So it turns out that `json.Decode()` internally generates a map of properties, so we can't even get the order because:

> ... the standard library's map type is implemented as a hash table, and iteration order (e.g., via range) is not guaranteed and can even be randomized across program runs for security reasons.

The fact that the unit tests were failing randomly is that the `map` _intentionally_ randomizes the order of fields returned from `range`.

PR #746, opened 2026-01-08
Add Concurrency Safety Documentation for Hooks

IDisposable:
Added documentation about not referencing Events after being written. Closes #733 documentation request.

Builds on #745

PR #747, opened 2026-01-08
Ensure that Event and Array objects returned to pool are cleared

IDisposable:
Help ensure no corruptible state remains in objects being returned to the pool.
Always clear the contextual properties to ensure the recycled (or erroneously referenced)
Event or Array are not accidentally used/shared.
Sets the truncates the buffer (setting length, not capacity, to zero).
Make it really obvious when you hold references as mentioned in #733
Clarified the documentation about holding references to CreateDict() events.

Builds on #745 and #746


IDisposable:
@rs this (and it's predecessors) are ready.

Issue #591, opened 2023-09-28
Add a dedicated Hook for Fatal calls


LorisFriedel:
Hello!

I'm currently converting all my projects from logrus to zerolog (thank you guys for the amazing library!) and I was wondering if it would be possible to add a dedicated Hook associated to Fatal call, something that would help cleanup things properly if a Fatal call happens.

Logrus has https://github.com/sirupsen/logrus/blob/master/alt_exit.go#L74 (DeferExitHandler) and it's very handy for when the Go program has CLI interface that manipulate the terminal in such way a cleanup is needed if the program exit for example.

I know I can do it with the Hook system already, but I feel like this special use-case would:
 1. improve performances compared to setting-up a classic Hook
 2. make sense to have a dedicated hook because of the nature of the Fatal call
 3. be easier to use and setup overall for this kind of use-cases

What do you think?

roachadam:
+1 for this. I'm struggling to get the zerolog hook to behave in the same manner as needed from logrus.

I can get it to work with normal logs, but using Fatal/Panic does not seem to pass the event back.

Issue #473, opened 2022-09-08
Fatal() does os.Exit(1) when log is disabled


lbergesio:
I think this is an error, It was discovered after upgrading from 1.26.1 to 1.28.0:

`Fatal()` here https://github.com/rs/zerolog/blob/v1.28.0/log.go#L367 calls `newEvent` (https://github.com/rs/zerolog/blob/d894f123bc5c2a887c95e90218b9410563141d67/log.go#L444) with the `done` paremeter being `os.Exit(1)` so here even the log is disabled done is executed which was not like this before:

https://github.com/rs/zerolog/blob/d894f123bc5c2a887c95e90218b9410563141d67/log.go#L448

lbergesio:
This has been introduced in 1.27.0

rs:
This is the expected behavior though. Disabling logging should only disable the logging and not change the flow of you code.

lbergesio:
The problem to me is that the logger I am using is Nop() and it reads: Nop returns a disabled logger for which all operation are no-op. So I would expect that even Fatal() is not op for it, meaning not exiting my program. Does it make sense?

rs:
If you call Fatal on a logger in your code, you are not supposed to know the logger is disabled, and the code logic is not supposed to pass the Fatal statement.

lbergesio:
Agree. My point is more in the semantics of "no op" not if in the log is disabled or not. But if according to you a  nop logger should be exiting ok, i will just modify my code that was using 1.26.1. this can be closed in such case. Thanks

PR #748, opened 2026-01-12
Test coverage improvements

IDisposable:
Implements #397 and #591, might help with #473
Test coverage for core is 97.6% with only real fringe cases remaining.
Improve `Fields` `isNilValue()` portability.
Added a new global handler `FatalExitFunc` to allow intercepting `Fatal()` messages.
Fixed CBOR float constants (removed the CBOR prefix)

Added tests for:
  - global `FatalExitFunc` to allow intercepting Fatal messages (both for testing and public use).
  - `Logger`
    -  `DisableSampling`
    -  `.With()` copying existing context if present.
    -  `.With().Fields()` all forms of `ErrorStackMarshaler` returns.
    - `.WithLevel()` for `FatalLevel`, `PanicLevel`, and `DisabledLevel`.
    - `.Err()` with `nil` and non-`nil` `error` and test the resulting log level.
    - `.should()` covering `nil` writer.
    - `.Output()` gets context values.
    - `.UpdateContext()` on a disabled logger doesn't panic and is a nop.
    - `.With()` all forms of `ErrorStackMarshaler` returns.
    - with a `nil`writer.
  - `.Hook()` passing no hooks.
 - `Array`
    - `.MarshalZerologArray` is a nop that won't panic.
  - `Context`
    - ` .Err()` and `.AnErr()` for `nil` errors and all forms of `ErrorStackMarshaler` returns.
  - `Event`
    - `.Caller()` to ensure we don't panic or add invalid information if `runtime.Caller()` fails
    - ` .Err()` and `.AnErr()` for `nil` errors and all forms of `ErrorStackMarshaler` returns.
  - `Fields`
    - ` .appendFields()`  all forms of `ErrorStackMarshaler` returns.
  - `HookLevel`
    - `.Run()` methods.
  - `LevelSampler`
    - `.Sample() methods.
  - `Syslog`
    - `.Write()`, `.WriteLevel()`, and `.Close()` methods.
    - `.WriteLevel()` with an `InvalidLevel`.
  - `Writer`
    - `.Write()` short write and error cases.
    - `MultiLevelWriter` `.WriteLevel()` and for `.Write()` error and  `.Close()` cases.
  - test of unmarshalling a level byte returns correct error.
  - CBOR decodeStream
    - `.decodeFloat()`, `.binaryFmt()`, `.DecodeIfBinaryToString()`, `.DecodeObjectToStr()`, `.DecodeIfBinaryToBytes()`, `.decodeTagData()`, and `.decodeSimpleFloat()`
    - handling of invalid UTF-8 sequences
    - handling of UTC times.
    - handling of timestamps
    - handling of various map lengths

Restructure `Event` `.caller()` so we test for ok and eliminate untestable coverage hole.

Restructure `Context` `.Err()` when the `ErrorStackMarshaler` returns a `nil` so there's code to cover.

Inverted logic for`Event` `.Caller()`'s call to `runtime.Caller()` for simpler testing.

Inverted logic for `Array` `.putArray()` and `Event` `.putEvent()` so there isn't uncoverable code.

Restructure `Field` `.appendFieldList()` to early return when `ErrorStackMarshaler` returns a `nil`

Added comments for things we can't get coverage on.

Did a go fmt ./...

Coverage of core is now 100% on `Array`, `Context`, `Ctx`, `Event`, `Field`, `Hook`, and `Syslog`.

Coverage of `Globals`, `Log`, `Sampler`, and `Writer` is almost all except some real edge-cases.

JSON encoder coverage is 100%

CBOR encoder coverage is 96.3% with base, cbor, string, time and types at 100% and decode_stream (which it lacks coverage on some panic states, and two incorrect coverage-tool lapses)

### DIFF
--- a/context_test.go
+++ b/context_test.go
@@ -25,7 +25,7 @@ func TestContext_ErrWithStackMarshaler(t *testing.T) {
 
 	log.Info().Msg("test message")
 
-	got := buf.String()
+	got := decodeIfBinaryToString(buf.Bytes())
 	want := `{"level":"info","stack":"stack-trace","error":"test error","message":"test message"}` + "\n"
 	if got != want {
 		t.Errorf("Context.Err() with stack marshaler = %q, want %q", got, want)
@@ -47,7 +47,7 @@ func TestContext_AnErrWithNilErrorMarshal(t *testing.T) {
 
 	log.Info().Msg("test message")
 
-	got := buf.String()
+	got := decodeIfBinaryToString(buf.Bytes())
 	want := `{"level":"info","message":"test message"}` + "\n" // No "test" field because isNilValue returned true
 	if got != want {
 		t.Errorf("Context.AnErr() with nil error marshal = %q, want %q", got, want)
@@ -69,7 +69,7 @@ func TestContext_ErrWithNilStackMarshaler(t *testing.T) {
 
 	log.Info().Msg("test message")
 
-	got := buf.String()
+	got := decodeIfBinaryToString(buf.Bytes())
 	want := `{"level":"info","message":"test message"}` + "\n" // No stack or error field because stack marshaler returned nil
 	if got != want {
 		t.Errorf("Context.Err() with nil stack marshaler = %q, want %q", got, want)
@@ -91,7 +91,7 @@ func TestContext_ErrWithStackMarshalerObject(t *testing.T) {
 
 	log.Info().Msg("test message")
 
-	got := buf.String()
+	got := decodeIfBinaryToString(buf.Bytes())
 	want := `{"level":"info","stack":{"name":"user","age":-30},"error":"test error","message":"test message"}` + "\n"
 	if got != want {
 		t.Errorf("Context.Err() with stack marshaler object = %q, want %q", got, want)
@@ -113,7 +113,7 @@ func TestContext_ErrWithStackMarshalerError(t *testing.T) {
 
 	log.Info().Msg("test message")
 
-	got := buf.String()
+	got := decodeIfBinaryToString(buf.Bytes())
 	want := `{"level":"info","stack":"stack error","error":"test error","message":"test message"}` + "\n"
 	if got != want {
 		t.Errorf("Context.Err() with stack marshaler error = %q, want %q", got, want)
@@ -135,7 +135,7 @@ func TestContext_ErrWithStackMarshalerInterface(t *testing.T) {
 
 	log.Info().Msg("test message")
 
-	got := buf.String()
+	got := decodeIfBinaryToString(buf.Bytes())
 	want := `{"level":"info","stack":42,"error":"test error","message":"test message"}` + "\n"
 	if got != want {
 		t.Errorf("Context.Err() with stack marshaler interface = %q, want %q", got, want)


internal/cbor/cbor.go
```diff
@@ -55,12 +55,12 @@ const (
 )
 
 const (
-	float32Nan         = "\xfa\x7f\xc0\x00\x00"
```
IDisposable, line 58: These constants included the CBOR prefixes instead of just the float's encoding. They were only being referenced in the `decoder_test.go` where we really needed them without the type prefix for other coverage tests.

array.go
```diff
@@ -38,10 +38,9 @@ func putArray(a *Array) {
 	//
 	// See https://golang.org/issue/23199
 	const maxSize = 1 << 16 // 64KiB
-	if cap(a.buf) > maxSize {
-		return
+	if cap(a.buf) <= maxSize {
```
IDisposable, line 41: By inverting the logic for this, there's nothing we can't test (there was a coverage hole here)

array.go
```diff
@@ -60,6 +59,7 @@ func Arr() *Array {
 // MarshalZerologArray method here is no-op - since data is
 // already in the needed format.
 func (*Array) MarshalZerologArray(*Array) {
+	// untestable: there's no code to be covered
```
IDisposable, line 62: Just a note so the reason for the coverage whole it noted.

console.go
```diff
@@ -101,9 +101,9 @@ type ConsoleWriter struct {
 // NewConsoleWriter creates and initializes a new ConsoleWriter.
 func NewConsoleWriter(options ...func(w *ConsoleWriter)) ConsoleWriter {
 	w := ConsoleWriter{
-		Out:          os.Stdout,
-		TimeFormat:   consoleDefaultTimeFormat,
-		PartsOrder:   consoleDefaultPartsOrder(),
+		Out:        os.Stdout,
```
IDisposable, line 104: Just a long-overdue `go fmt ./...`

context.go
```diff
@@ -184,7 +184,7 @@ func (c Context) Err(err error) Context {
 	if c.l.stack && ErrorStackMarshaler != nil {
 		switch m := ErrorStackMarshaler(err).(type) {
 		case nil:
-			// do nothing
+			return c // do nothing with nil errors
```
IDisposable, line 187: By adding an explicit return here there is something for the code-coverage to observe.

ctx.go
```diff
@@ -25,12 +25,11 @@ type ctxKey struct{}
 // replacing it in a new Context), use UpdateContext with the following
 // notation:
 //
-//     ctx := r.Context()
-//     l := zerolog.Ctx(ctx)
-//     l.UpdateContext(func(c Context) Context {
-//         return c.Str("bar", "baz")
-//     })
-//
+//	ctx := r.Context()
```
IDisposable, line 28: `go fmt ./...`

event.go
```diff
@@ -48,10 +48,9 @@ func putEvent(e *Event) {
 	//
 	// See https://golang.org/issue/23199
 	const maxSize = 1 << 16 // 64KiB
-	if cap(e.buf) > maxSize {
-		return
+	if cap(e.buf) <= maxSize {
```
IDisposable, line 51: By inverting the logic for this, there's nothing we can't test (there was a coverage hole here)

event.go
```diff
@@ -435,7 +434,7 @@ func (e *Event) Err(err error) *Event {
 	if e.stack && ErrorStackMarshaler != nil {
 		switch m := ErrorStackMarshaler(err).(type) {
 		case nil:
-			// do nothing
+			return e
```
IDisposable, line 437: By adding an explicit return here there is something for the code-coverage to observe.

event.go
```diff
@@ -835,11 +834,9 @@ func (e *Event) caller(skip int) *Event {
 	if e == nil {
 		return e
 	}
-	pc, file, line, ok := runtime.Caller(skip + e.skipFrame)
-	if !ok {
-		return e
+	if pc, file, line, ok := runtime.Caller(skip + e.skipFrame); ok {
```
IDisposable, line 837: By inverting the logic for this, there's nothing we can't test (there was a coverage hole here)

fields.go
```diff
@@ -5,13 +5,18 @@ import (
...
 	"net"
+	"reflect"
 	"sort"
 	"time"
-	"unsafe"
 )
 
-func isNilValue(i interface{}) bool {
-	return (*[2]uintptr)(unsafe.Pointer(&i))[1] == 0
+func isNilValue(e error) bool {
```
IDisposable, line 13: This seems to be a more-portable way to express this check, and unit tests can show coverage.

fields.go
```diff
@@ -77,7 +82,7 @@ func appendFieldList(dst []byte, kvList []interface{}, stack bool, ctx context.C
 			if stack && ErrorStackMarshaler != nil {
 				switch m := ErrorStackMarshaler(val).(type) {
 				case nil:
-					// do nothing
+					return dst // do nothing with nil errors
```
IDisposable, line 85: By adding an explicit return here there is something for the code-coverage to observe.

globals.go
```diff
@@ -134,6 +134,10 @@ var (
 	// be thread safe and non-blocking.
 	ErrorHandler func(err error)
 
+	// FatalExitFunc is called by log.Fatal() instead of os.Exit(1). If not set,
```
IDisposable, line 137: New feature (needed for testing `Fatal()` that also addresses a couple outstanding issues #397 and #591, and might help with #473

log.go
```diff
@@ -2,85 +2,85 @@
 //
 // A global Logger can be use for simple logging:
 //
-//     import "github.com/rs/zerolog/log"
+//	import "github.com/rs/zerolog/log"
```
IDisposable, line 5: `go fmt ./...`

log.go
```diff
@@ -382,18 +382,24 @@ func (l *Logger) Err(err error) *Event {
...
 	return l.newEvent(FatalLevel, func(msg string) {
 		if closer, ok := l.w.(io.Closer); ok {
 			// Close the writer to flush any buffered message. Otherwise the message
-			// will be lost as os.Exit() terminates the program immediately.
+			// could be lost if FatalExitFunc() terminates the program immediately or
+			// os.Exit(1) is called if not FatalExitFunc isn't set (default).
 			closer.Close()
 		}
-		os.Exit(1)
+		if FatalExitFunc != nil {
```
IDisposable, line 398: New feature, a hook to replace/handle the `Fatal()` `os.Exit(1)` so you can test, or even change the exit value.

fields.go
```diff
@@ -5,13 +5,18 @@ import (
...
-func isNilValue(i interface{}) bool {
-	return (*[2]uintptr)(unsafe.Pointer(&i))[1] == 0
+func isNilValue(e error) bool {
+	switch reflect.TypeOf(e).Kind() {
+	case reflect.Ptr:
+		return reflect.ValueOf(e).IsNil()
+	default:
+		return false
+	}
 }
```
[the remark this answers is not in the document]
IDisposable: That's how I originally wrote it, but then I can't get code coverage because that path _never_ occurs. Not going to change this

context_test.go
```diff
@@ -0,0 +1,143 @@
...
+	ErrorStackMarshaler = func(err error) interface{} {
+		return "stack-trace"
+	}
+
+	var buf bytes.Buffer
+	log := New(&buf).With().Stack().Err(errors.New("test error")).Logger()
+
+	log.Info().Msg("test message")
+
+	got := decodeIfBinaryToString(buf.Bytes())
```
IDisposable, line 28: _note to self_ Remember to run `go test ./... -tags binary_log` when checking coverage!

internal/cbor/string_test.go
```diff
@@ -96,7 +97,7 @@ func TestAppendStrings(t *testing.T) {
 		array = append(array, tt.plain)
 	}
 	want := make([]byte, 0)
-	want = append(want, 0x94) // start array length 24
+	want = append(want, 0x95) // start array
```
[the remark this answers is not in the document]
IDisposable: That comment is completely incorrect. 0x94 is type-code 4 (array) with length of 21 (not 4)

fields.go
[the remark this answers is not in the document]
IDisposable: That path cannot be called with an untyped `nil`, it will always be a non-`nil` `error` and if I place a check in there, it is unreachable in any test code. Not needed.

log_test.go
```diff
@@ -795,15 +859,168 @@ func TestLevelWriter(t *testing.T) {
...
+	}
+
+	// Allow extra-verbose logs.
+	SetGlobalLevel(TraceLevel - 1)
+	log := New(lw).Level(TraceLevel - 1)
+
+	// Set FatalExitFunc to panic so we can catch it
+	oldFatalExitFunc := FatalExitFunc
+	FatalExitFunc = func() { panic("fatal exit") }
+	defer func() { FatalExitFunc = oldFatalExitFunc }()
```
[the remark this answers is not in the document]
IDisposable: This is a technically correct, but completely irrelevant change.

PR #754, opened 2026-02-10
fix: UpdateContext skips Nop and zero-value loggers

### LINKED ISSUE #643: logger.UpdateContext checks for disabled logger using equality
`logger.UpdateContext` checks for disabled logger:

```
	if l == disabledLogger {
		return
	}
```

I think this is problematic as it should check in a more general way, to handle also `Nop` and zero values. In my tests I found race conditions because I was reusing same global nop-logger. I think a better check would be something like it was done in #617. Do you agree? I can make a PR.

veeceey:
## Summary

Fixes #643.

`UpdateContext` used pointer equality (`l == disabledLogger`) to detect disabled loggers. This only caught the package-level `disabledLogger` singleton returned by `Ctx()` when no logger is in the context, but **missed** loggers created via:
- `zerolog.Nop()` (returns a new disabled logger each time)
- Zero-value `Logger{}` (has nil writer)

When a shared `Nop()` logger was used from multiple goroutines, `UpdateContext` would proceed to mutate the context buffer concurrently, causing **data races** (confirmed with `-race` detector, reproduces 15 race warnings).

## Changes

- **`log.go`**: Replace `l == disabledLogger` with `l.w == nil || l.level == Disabled`, consistent with how `Logger.should()` determines if logging is active
- **`log_test.go`**: Add `TestUpdateContextOnNopLogger` and `TestUpdateContextOnZeroValueLogger` to cover both missed cases

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] New tests verify `UpdateContext` is a no-op on `Nop()` loggers
- [x] New tests verify `UpdateContext` is a no-op on zero-value loggers
- [x] Data race no longer reproduces with `go run -race` on shared `Nop()` logger

### DIFF
--- a/log.go
+++ b/log.go
@@ -294,7 +294,7 @@ func (l Logger) With() Context {
 // Caution: This method is not concurrency safe.
 // Use the With method to create a child logger before modifying the context from concurrent goroutines.
 func (l *Logger) UpdateContext(update func(c Context) Context) {
-	if l.w == nil || l.level == Disabled {
+	if l.disabled() {
 		return
 	}
 	if cap(l.context) == 0 {
@@ -508,9 +508,14 @@ func (l *Logger) scratchEvent() *Event {
 	return newEvent(LevelWriterAdapter{io.Discard}, DebugLevel, l.stack, l.ctx, l.hooks)
 }
 
+// disabled returns true if the logger is a disabled or nop logger.
+func (l *Logger) disabled() bool {
+	return l.w == nil || l.level == Disabled
+}
+
 // should returns true if the log event should be logged.
 func (l *Logger) should(lvl Level) bool {
-	if l.w == nil {
+	if l.disabled() {
 		return false
 	}
 	if lvl < l.level || lvl < GlobalLevel() {


log.go
```diff
@@ -294,7 +294,7 @@ func (l Logger) With() Context {
 // Caution: This method is not concurrency safe.
 // Use the With method to create a child logger before modifying the context from concurrent goroutines.
 func (l *Logger) UpdateContext(update func(c Context) Context) {
-	if l == disabledLogger {
+	if l.w == nil || l.level == Disabled {
```
IDisposable: Might be better to add a method to _log.go_ that encapsulates this and call here and in `should()`...

```go
// detects if this is a disabled logger
func (l *Logger) disabled() bool {
   return l.w == nil || l.level <= Disabled
}
```

then here we do 

```go
if l.disabled() {
   return
}
```

and then _should()_ starts with 

```go
// should returns true if the log event should be logged.
func (l *Logger) should(lvl Level) bool {
	if l.disabled() {
		return false
	}
...
```

veeceey: Good call — done. Extracted a `disabled()` method and updated both `UpdateContext` and `should()` to use it.

IDisposable (approved):
:shipit:

log.go
```diff
@@ -508,9 +508,14 @@ func (l *Logger) scratchEvent() *Event {
 	return newEvent(LevelWriterAdapter{io.Discard}, DebugLevel, l.stack, l.ctx, l.hooks)
 }
 
+// disabled returns true if the logger is a disabled or nop logger.
+func (l *Logger) disabled() bool {
+	return l.w == nil || l.level == Disabled
```
mitar, line 513: I think this check should be:


```suggestion
	return l.w == nil || l.level == Disabled || GlobalLevel() == Disabled
```

This is why I made and tested in https://github.com/rs/zerolog/pull/617/changes.

PR #755, opened 2026-02-17
Add slog.Handler implementation for zerolog

### LINKED ISSUE #571: stdlib slog handler for zerolog

veeceey:
Closes #571

I needed slog interop in a project where zerolog handles all log output, but some dependencies use `log/slog`. Rather than losing zerolog's performance by switching to slog's built-in JSON handler, this adds a `SlogHandler` that implements `slog.Handler` and routes everything through zerolog.

### What this does

`zerolog.NewSlogHandler(logger)` returns a `slog.Handler` backed by the given `zerolog.Logger`. You can use it like:

```go
zl := zerolog.New(os.Stderr).With().Timestamp().Logger()
slog.SetDefault(slog.New(zerolog.NewSlogHandler(zl)))

slog.Info("request handled", "method", "GET", "status", 200)
// Output: {"level":"info","method":"GET","status":200,"time":...,"message":"request handled"}
```

**Level mapping:**
- `slog.LevelDebug-4` and below -> `zerolog.TraceLevel`
- `slog.LevelDebug` -> `zerolog.DebugLevel`
- `slog.LevelInfo` -> `zerolog.InfoLevel`
- `slog.LevelWarn` -> `zerolog.WarnLevel`
- `slog.LevelError` -> `zerolog.ErrorLevel`

**Supported features:**
- All slog attribute types encoded with zerolog's typed methods (no reflection for primitives)
- `WithAttrs` for pre-attaching fields to child handlers
- `WithGroup` for namespacing keys with dot-separated prefixes
- Nested groups work correctly
- `LogValuer` resolution
- Level filtering respects the zerolog Logger's configured level
- Zerolog contextual fields from `With()` are preserved

**Files:**
- `slog.go` - the handler implementation (~200 lines)
- `slog_test.go` - 26 tests covering levels, all attr types, groups, filtering, LogValuer, immutability

All existing tests continue to pass (the `RandomSampler` flake in `sampler_test.go` is pre-existing).

The `go.mod` already requires Go 1.23, so `log/slog` is available without any changes.

### DIFF
--- a/slog.go
+++ b/slog.go
@@ -3,7 +3,6 @@ package zerolog
 import (
 	"context"
 	"log/slog"
-	"strings"
 	"time"
 )
 
@@ -24,17 +23,31 @@ func NewSlogHandler(logger Logger) *SlogHandler {
 }
 
 // Enabled reports whether the handler handles records at the given level.
+// It mirrors Logger.should's level and writer checks (without sampling).
 func (h *SlogHandler) Enabled(_ context.Context, level slog.Level) bool {
-	return h.logger.GetLevel() <= slogToZerologLevel(level)
+	if h.logger.w == nil {
+		return false
+	}
+	zl := slogToZerologLevel(level)
+	if zl < GlobalLevel() {
+		return false
+	}
+	return zl >= h.logger.level
 }
 
 // Handle handles the Record. It converts the slog.Record into a zerolog event
 // and writes it using the underlying zerolog.Logger.
-func (h *SlogHandler) Handle(_ context.Context, record slog.Record) error {
+func (h *SlogHandler) Handle(ctx context.Context, record slog.Record) error {
 	zlevel := slogToZerologLevel(record.Level)
 	event := h.logger.WithLevel(zlevel)
 	if event == nil {
 		return nil
+	}
+
+	// Propagate slog context to the zerolog event so that hooks
+	// relying on Event.GetCtx() (e.g. tracing) can access it.
+	if ctx != nil {
+		event = event.Ctx(ctx)
 	}
 
 	// Add pre-attached attrs from WithAttrs
@@ -48,13 +61,26 @@ func (h *SlogHandler) Handle(_ context.Context, record slog.Record) error {
 		return true
 	})
 
-	// Add timestamp
-	if !record.Time.IsZero() {
+	// Add timestamp from the slog record, but only if the logger doesn't
+	// already have a timestampHook (added via .With().Timestamp()) to
+	// avoid duplicate timestamp keys in the output.
+	if !record.Time.IsZero() && !h.hasTimestampHook() {
 		event.Time(TimestampFieldName, record.Time)
 	}
 
 	event.Msg(record.Message)
 	return nil
+}
+
+// hasTimestampHook reports whether the logger has a timestampHook installed,
+// which would cause duplicate timestamp fields if we also emit record.Time.
+func (h *SlogHandler) hasTimestampHook() bool {
+	for _, hook := range h.logger.hooks {
+		if _, ok := hook.(timestampHook); ok {
+			return true
+		}
+	}
+	return false
 }
 
 // WithAttrs returns a new Handler with the given attributes pre-attached.
@@ -136,10 +162,16 @@ func zerologToSlogLevel(level Level) slog.Level {
 	}
 }
 
-// joinPrefix concatenates a prefix and key with a dot separator,
-// trimming any leading or trailing dots from the result.
+// joinPrefix concatenates a prefix and key with a dot separator.
+// It avoids allocations when either prefix or key is empty.
 func joinPrefix(prefix, key string) string {
-	return strings.Trim(prefix+"."+key, ".")
+	if prefix == "" {
+		return key
+	}
+	if key == "" {
+		return prefix
+	}
+	return prefix + "." + key
 }
 
 // appendSlogAttr appends a single slog.Attr to the zerolog event, handling
--- a/slog_test.go
+++ b/slog_test.go
@@ -1,6 +1,7 @@
 package zerolog_test
 
 import (
+	"context"
 	"bytes"
 	"encoding/json"
 	"errors"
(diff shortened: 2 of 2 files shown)


slog.go
```diff
@@ -0,0 +1,207 @@
...
+// type-specific encoding to avoid reflection where possible.
+func appendSlogAttr(event *Event, attr slog.Attr, prefix string) *Event {
+	if event == nil {
+		return event
+	}
+
+	// Resolve the attribute to handle LogValuer types
+	attr.Value = attr.Value.Resolve()
+
+	key := attr.Key
```
IDisposable: Since this is a called in a really tight loop for all the attributes, might be better to reduce the branch/potential for execution stalls.  Do we know if prefix is _always_ non-blank here? If so, we could just always do the concatenation.

slog.go
```diff
@@ -0,0 +1,207 @@
...
+	case slog.KindTime:
+		event = event.Time(key, val.Time())
+	case slog.KindGroup:
+		attrs := val.Group()
+		if len(attrs) == 0 {
+			return event
+		}
+		groupPrefix := prefix
+		if key != "" {
+			groupPrefix = key
```
IDisposable: Seems confusing that by the time we get to this line, `prefix` is what we were passed in, and key has already concatenated the prefix and a dot to the key (unless key was blank)... so essentially this code is _discarding_ the already concatenated string). I wonder two things:

1) we've already had to test `attr.Value.Kind() != slog.KindGroup` up above so maybe this entire block should be moved up there
2) could we just always concatenate the `prefix + "." + key` (and slice off the bare trailing `.` if the `key was blank?)

slog.go
```diff
@@ -0,0 +1,207 @@
...
+		switch cv := v.(type) {
+		case error:
+			event = event.AnErr(key, cv)
+		case time.Duration:
+			event = event.Dur(key, cv)
+		case time.Time:
+			event = event.Time(key, cv)
+		case []byte:
+			event = event.Bytes(key, cv)
+		default:
```
IDisposable, line 236: Should we also be handling `KindLogValuer`?

slog.go
veeceey: Good point about the tight loop. Prefix is most commonly non-blank when we're inside nested groups, but can be empty at the top level. I think unconditional concatenation makes sense here - we can always concatenate `prefix + "." + key` and then just use a helper to clean up leading/trailing dots if needed. Eliminates the branch and makes the code simpler.

slog.go
veeceey: Yeah, I see what you mean - we're already checking the kind above so it is a bit redundant. And you're right about the key concatenation getting discarded in the group case. I like the idea of moving the entire prefix logic up higher and just always concatenating `prefix + "." + key`, then stripping leading/trailing dots if needed. That would be cleaner and more efficient.

slog.go
veeceey: Good catch - actually we do handle it implicitly since we call `attr.Value.Resolve()` at the start which resolves LogValuer implementations. But it might be worth being more explicit about it. Let me look at whether we should add a specific case for it or if the current approach is sufficient.

slog.go
veeceey: Pushed the changes - extracted a `joinPrefix` helper that just does `strings.Trim(prefix+"."+key, ".")` unconditionally. No more branching in the hot path.

slog.go
veeceey: Agreed, moved the group handling up before key concatenation so it's handled first. The group case now uses `joinPrefix(prefix, attr.Key)` directly for its groupPrefix, and non-group attrs skip early if key is empty. Much cleaner flow now.

slog.go
veeceey: Added a more explicit comment clarifying that `attr.Value.Resolve()` handles KindLogValuer by unwrapping it. Since Resolve() is called before the switch, we don't need a separate case for it - it'll already be resolved to its underlying kind by that point.

veeceey:
addressed all three review points — extracted a joinPrefix helper to clean up the branching, moved group handling before key concat, and added a note about LogValuer resolution. pushed the updates, lmk if anything else needs adjusting

slog.go
```diff
@@ -0,0 +1,215 @@
+package zerolog
```
rs, line 1: Does it need to be in the same package?

slog.go
```diff
@@ -0,0 +1,215 @@
...
+// NewSlogHandler creates a new slog.Handler that writes log records to the
+// given zerolog.Logger. The handler maps slog levels to zerolog levels and
+// converts slog attributes to zerolog fields.
+func NewSlogHandler(logger Logger) *SlogHandler {
+	return &SlogHandler{logger: logger}
+}
+
+// Enabled reports whether the handler handles records at the given level.
+func (h *SlogHandler) Enabled(_ context.Context, level slog.Level) bool {
+	return h.logger.GetLevel() <= slogToZerologLevel(level)
```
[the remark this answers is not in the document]
veeceey: Fixed - Enabled() now checks h.logger.w \!= nil and compares against GlobalLevel() in addition to the logger's own level, matching Logger.should's behavior (minus sampling).

slog.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// Handle handles the Record. It converts the slog.Record into a zerolog event
+// and writes it using the underlying zerolog.Logger.
+func (h *SlogHandler) Handle(_ context.Context, record slog.Record) error {
+	zlevel := slogToZerologLevel(record.Level)
+	event := h.logger.WithLevel(zlevel)
+	if event == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
veeceey: Fixed - Handle() now propagates the slog ctx to the zerolog event via event.Ctx(ctx), so hooks that call Event.GetCtx() will see it.

slog.go
```diff
@@ -0,0 +1,215 @@
...
+	record.Attrs(func(a slog.Attr) bool {
+		event = appendSlogAttr(event, a, h.prefix)
+		return true
+	})
+
+	// Add timestamp
+	if !record.Time.IsZero() {
+		event.Time(TimestampFieldName, record.Time)
+	}
+
```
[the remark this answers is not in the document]
veeceey: Fixed - Handle() now checks whether the logger has a timestampHook installed (via hasTimestampHook()) and skips emitting record.Time when it does. If the logger doesn't have the hook, it still writes the slog record's timestamp as before.

slog.go
```diff
@@ -0,0 +1,215 @@
...
+	case PanicLevel:
+		return slog.LevelError + 8
+	default:
+		return slog.LevelInfo
+	}
+}
+
+// joinPrefix concatenates a prefix and key with a dot separator,
+// trimming any leading or trailing dots from the result.
+func joinPrefix(prefix, key string) string {
```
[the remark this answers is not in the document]
veeceey: Fixed - joinPrefix now short-circuits: returns key directly when prefix is empty, returns prefix when key is empty, only concatenates when both are non-empty. Also dropped the strings.Trim call since the inputs are clean.

slog.go
IDisposable: We could require it be opted in with a tag (like we do for CBOR), or if we want to move it out, we would need to expose the helpers it needs in the public API that it can do the needful

veeceey:
That's a fair concern. A build tag for opt-in makes sense to me — keeps it zero-cost for users who don't need slog compatibility. I'll restructure it behind a tag. Do you have a preference on the tag name? Something like `slog` or `zerolog_slog`?

PR #763, opened 2026-04-13
event: restore Err() logging when ErrorStackMarshaler returns nil

### LINKED ISSUE #762: Behavior change on `Event.Err` with commit f6fbd33
Hello!

I upgraded from zerolog v1.31.0 to v1.35.0 and some of my tests validating log format started failing. I use the https://github.com/elastic/ecs-logging-go-zerolog library to enforce a formatting standard and saw that the error messages stopped appearing.

I tracked the issue to this commit:
https://github.com/rs/zerolog/commit/f6fbd330be6e308d528e1f0bd7972c358b41370f#diff-552f47512a00afe5fc6850cc9ddc830a6daeca162750e50aab4ed549685e0253R187

Before, when `ErrorStackMarshaller(err).(type)` returned `nil` we would fall through the switch case and execute `return e.AnErr(ErrorFieldName, err)`. This ensured that a regular error without stack wrapping would still get marshalled.

After the commit, the function started returning the event unchanged, as if the error was `nil` which is not the case. How should I deal with this behaviour change?

alliasgher:
Fixes #762. Commit f6fbd33 introduced a 'case nil: return e' that drops the error field entirely when ErrorStackMarshaler returns nil. This restores the original behaviour of always calling AnErr even when no stack is attached.

### DIFF
--- a/event.go
+++ b/event.go
@@ -466,7 +466,8 @@ func (e *Event) Err(err error) *Event {
 	if e.stack && ErrorStackMarshaler != nil {
 		switch m := ErrorStackMarshaler(err).(type) {
 		case nil:
-			return e
+			// ErrorStackMarshaler returned nil — the error has no stack trace to
+			// attach. Fall through and still log the error via AnErr below.
 		case LogObjectMarshaler:
 			e = e.Object(ErrorStackFieldName, m)
 		case error:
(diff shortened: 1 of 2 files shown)


IDisposable (approved):

IDisposable:
This one only reverts one of the changes, probably should prefer #758

PR #764, opened 2026-04-15
fix: Err() should record error when ErrorStackMarshaler returns nil

### LINKED ISSUE #762: Behavior change on `Event.Err` with commit f6fbd33

Yanhu007:
Fixes #762

## Problem

Since commit f6fbd33, when `ErrorStackMarshaler` returns `nil` (e.g. for errors that don't have stack traces), `Err()` returns immediately without calling `AnErr(ErrorFieldName, err)`. This silently drops the error message from log output.

Before the regression:
```json
{"level":"error","error":"something failed","message":"..."}
```

After the regression:
```json
{"level":"error","message":"..."}
```
(error field is missing!)

## Fix

Change `case nil: return e` to an empty `case nil:` so control falls through to the `AnErr()` call that records the error field. The stack field is correctly skipped when the marshaler returns nil.

### DIFF
--- a/context.go
+++ b/context.go
@@ -210,7 +210,8 @@ func (c Context) Err(err error) Context {
 	if c.l.stack && ErrorStackMarshaler != nil {
 		switch m := ErrorStackMarshaler(err).(type) {
 		case nil:
-			return c // do nothing with nil errors
+			// Stack marshaler returned nil; skip the stack field
+			// but still record the error below.
 		case LogObjectMarshaler:
 			c = c.Object(ErrorStackFieldName, m)
 		case error:
--- a/event.go
+++ b/event.go
@@ -466,7 +466,8 @@ func (e *Event) Err(err error) *Event {
 	if e.stack && ErrorStackMarshaler != nil {
 		switch m := ErrorStackMarshaler(err).(type) {
 		case nil:
-			return e
+			// Stack marshaler returned nil; skip the stack field
+			// but still record the error below.
 		case LogObjectMarshaler:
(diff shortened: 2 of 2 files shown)


Yanhu007:
Good catch on `Context.Err()` having the same issue! I've pushed an additional commit that fixes the identical `case nil: return c` early return in `context.go` as well.

Both `Event.Err()` and `Context.Err()` now correctly skip only the stack field when the marshaler returns nil, while still recording the error itself.

IDisposable (approved):

IDisposable:
This one doesn't update the unit tests, probably should prefer #758

IDisposable:
#763 was merged so this should be closed.

PR #769, opened 2026-05-13
ctx: propagate Go context to events created via Ctx(ctx)

### LINKED ISSUE #756: Issue with log.Ctx and event context
There is a very confusing problem for me when using zerolog:

`log.Ctx(ctx)` doesn't insert the provided context in the Event, resulting in logs without telemetry data

The solution is quite ugly:

```go
import "github.com/rs/zerolog/log"

...

log.
  Ctx(ctx). // Retrieves the logger from context, without setting the context for Event.
  Info(). // Creates an event
  Ctx(ctx). // Sets the ctx for the Event so we can retrieve telemetry data.
  ...
```

I think it would be highly convenient that we do not have to set the event context if the log is already created from a context.
Am I doing something wrong here?

c-tonneslan:
Fixes #756.

Right now, when you store a logger in a context with `WithContext` and retrieve it later with `Ctx(ctx)`, events you create from that logger have `e.ctx == nil`. That means hooks can't access the request context without callers manually chaining `.Ctx(ctx)` on every single event, which kind of defeats the purpose of having context-aware logging.

The fix is straightforward: in `WithContext`, after calling `context.WithValue` to get the new context, set `l.ctx = newCtx` on the logger copy before returning. Because `l` is a value receiver, `&l` escapes to the heap, and the assignment after `context.WithValue` is visible through the stored pointer. No changes to `Ctx()` needed, so the `UpdateContext` pattern that `hlog` middleware relies on keeps working.

**Changes:**
- `ctx.go`: set `l.ctx` to the context returned by `WithContext` so events carry it
- `ctx_test.go`: add `TestCtxEventHasContext` to verify hooks see the context; update `TestCtx` and `TestCtxDisabled` to account for the `ctx` field now being set
- `hlog/hlog_test.go`: rewrite `TestNewHandler` to be behavior-based (the old `reflect.DeepEqual` check broke because `hlog` is a separate package and can't compare the unexported `ctx` field)

**Testing:**
```
go test ./...
```
All tests pass.

### DIFF
--- a/ctx.go
+++ b/ctx.go
@@ -35,7 +35,12 @@ func (l Logger) WithContext(ctx context.Context) context.Context {
 		// Do not store disabled logger.
 		return ctx
 	}
-	return context.WithValue(ctx, ctxKey{}, &l)
+	// Attach ctx to the logger so that events created via Ctx(ctx).Info() etc.
+	// automatically have the Go context set and available to hooks, without
+	// callers needing to chain .Ctx(ctx) manually on every event.
+	newCtx := context.WithValue(ctx, ctxKey{}, &l)
+	l.ctx = newCtx
+	return newCtx
 }
 
 // Ctx returns the Logger associated with the ctx. If no logger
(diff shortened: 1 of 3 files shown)


rs:
This is semantically incorrect to modify the receiver's context in a method name WithContext. It is not the expected behavior and this change may break existing code.

PR #783, opened 2026-07-23
fix(console): return error when ConsoleWriter.Out is nil

sonnemusk:
## Summary
A zero-value `ConsoleWriter` leaves `Out` nil. `Write` then panics when it tries to write the formatted line.

Return a clear error instead of panicking so misconfiguration fails safely.

## Test plan
- [x] `go test -run Console`
- [x] New unit test covering nil `Out`


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #793, opened 2026-08-24
chore: Isolate pkgerrors and drop pkg/errors from main module

aziret:
This PR removes github.com/pkg/errors as a dependency from the main part of the library and extracts dependency on github.com/pkg/errors to an external pkg

This PR also tries to address the issue [470](https://github.com/rs/zerolog/issues/470) from the main repo

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,6 +22,9 @@ jobs:
           ${{ runner.os }}-go-
     - name: Test
       run: go test -race -bench . -benchmem ./...
+    - name: Test pkgerrors
+      run: go test -race -bench . -benchmem ./...
+      working-directory: pkgerrors
     - name: Test CBOR
       run: go test -tags binary_log -race ./...
   coverage:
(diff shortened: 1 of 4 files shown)


pkg/pkgerrors/stacktrace.go
rs, line 1: why adding this extra file path level?

aziret: Usually external packages are located in /pkg

aziret: The main part of the library is no longer dependant on github.com/pkg/errors
Only users of `package pkgerrors` will have this transitive dependency on github.com/pkg/errors

rs: Yes, I understand that part. I’m just wondering why we couldn’t add the `go.mod` and `go.sum` files to the package's current location. I couldn't find any convention recommending a `pkg` directory for this use case.

aziret: I'm just used to moving external packages into /pkg and automatically did this 😄 

Reverted my changes back and added module in pkgerrors

PR #802, opened 2026-09-18
Encode typed-nil errors in Fields as null

jakezwang:
Encode typed-nil errors passed through `Fields` as `null`. Currently the key or slice separator is written, then the value is skipped, producing invalid JSON such as `{"error":,"errors":[,"middle",]}`. This affects both event and context fields.

Write a null value in the existing typed-nil branches, consistent with untyped nil errors. Regression tests cover scalar errors, error slices with nil elements around a non-nil error, and a following field through both Event and Context; they also run with CBOR encoding.

Validation: Go 1.23.12 and Go 1.24.13 JSON/CBOR race tests and `go vet ./...`; the regression fails before the fix. The full CI command `go test -race -bench . -benchmem ./...` also passes on Go 1.24.13.


Voyagerroc-Lab (approved):
Encoding typed-nil errors as null keeps JSON structurally valid and matches the existing untyped-nil behavior. The new cases cover event/context fields, nil elements in error slices, trailing fields, and both JSON/CBOR. The author reports race tests on Go 1.23/1.24 and go vet passing; the PR regression fails before the fix. Review performed with OpenAI Codex at the account owners request.

PR #803, opened 2026-09-20
fix(hlog): proxy http.Pusher through mutil.WrapWriter

### LINKED ISSUE #417: When using hlog.AccessHandler, HTTP2 pushing gets disabled
Because `w.(http.Pusher)` fails.

cipherprofessor:
Fixes #417

## Problem

`mutil.WrapWriter`'s own doc comment promises the returned proxy supports "the full method set of the proxied object," but it never probes for `http.Pusher` — so `w.(http.Pusher)` always fails on the returned proxy, even when the real underlying `http.ResponseWriter` supports HTTP/2 server push. This silently disables push for anything routed through `hlog.AccessHandler`.

The issue's suggested fix was swapping to `httpsnoop`, but `CONTRIBUTING.md` asks contributors to avoid new dependencies, and the project's stated philosophy is "lean and non-bloated" — so this instead extends the existing (already ad hoc) interface-probing pattern with a 5th probe, no new dependency, no refactor.

## Fix

`WrapWriter` already special-cases two shapes: the full `CloseNotifier+Flusher+Hijacker+ReaderFrom` set → `fancyWriter`, or just `Flusher` → `flushWriter` (anything else falls back to the minimal `basicWriter`, which already doesn't preserve every possible interface combination — that's pre-existing, unchanged scope). This adds `fancyPushWriter` and `flushPushWriter` for the same two branches when the underlying writer also implements `http.Pusher`.

Checked against `net/http`'s own HTTP/2 `ResponseWriter` implementation: it implements `Flusher`+`Pusher` but not `Hijacker`/`ReaderFrom`, so real HTTP/2 traffic hits `flushPushWriter`. `fancyPushWriter` is for a custom or composed `ResponseWriter` that happens to implement all five interfaces at once — net/http's own writers don't produce that combination.

A `ResponseWriter` that implements *only* `Pusher` (no `Flusher`) still falls through to `basicWriter` and loses it — same class of pre-existing incompleteness as e.g. "just `Hijacker`" today, not something this PR expands or narrows.

Worth noting: HTTP/2 server push itself is a declining browser feature (Chrome dropped it in 2022–23) — this fixes proxying correctness per `WrapWriter`'s documented contract, not chasing a hot feature.

## Testing

Added the package's first test file (`hlog/internal/mutil/writer_proxy_test.go`). Since this adds an extra embedding layer, I specifically verified method *promotion* still works correctly through it — not just that `Push` is newly reachable, but that `Flush`/`CloseNotify`/`Hijack`/`ReadFrom` still reach the underlying writer too (confirmed via a throwaway probe that `fancyPushWriter`/`flushPushWriter`, only ever constructed as pointers, correctly promote pointer-receiver methods from their value-embedded fields — the same mechanism the existing `fancyWriter`/`flushWriter` → `basicWriter` embedding already relies on one level down). Also added a negative test confirming the full "fancy" shape without `Pusher` still correctly does *not* claim `http.Pusher` support (unchanged behavior, previously untested).

`go build`, `go vet`, `gofmt`, and the full test suite are all clean.


rs:
The flushPushWriter is likely unnecessary. Ok to remove it?

cipherprofessor:
It's necessary, not redundant — verified this against Go's actual stdlib source before replying rather than just re-reading my own diff.

Go's own HTTP/2 `ResponseWriter` (`http2responseWriter` in `net/http/h2_bundle.go`) is exactly the shape `flushPushWriter` exists for: it implements `Flush`/`FlushError`, `CloseNotify`, and `Push`, but has no `Hijack` or `ReadFrom` methods at all (confirmed with `grep -n 'func (w \*http2responseWriter)' h2_bundle.go` against Go 1.25.5 — zero hits for either). That means a real HTTP/2 response, the primary real-world case this PR is fixing `Push` for, has `cn=true, fl=true, hj=false, rf=false, ps=true`. In `WrapWriter`, `hj`/`rf` being false skips the `fancyWriter`/`fancyPushWriter` branch entirely, and it falls into the `if fl { if ps { ... } }` branch — `flushPushWriter` specifically.

If `flushPushWriter` is removed and that branch just returns `&flushWriter{...}` regardless of `ps`, real HTTP/2 responses lose `Push` again — silently regressing the exact bug this PR fixes, for what's arguably the most common real-world caller.

It's also already covered by an existing test that would need to change: `TestWrapWriterPreservesPusher_FlushOnly` in `writer_proxy_test.go` constructs a `flushPusher` (Flush + Push only, no CloseNotifier/Hijacker/ReaderFrom) specifically to model this shape and asserts `Push` forwards correctly through it.

Happy to add a comment citing the stdlib method set directly if that'd make the reasoning clearer in the code itself — let me know.

rs:
Thinking more about it, I don’t think we should add support for http.Pusher here. HTTP/2 server push is effectively obsolete in modern clients, including browsers, and Go clients don’t support receiving pushes either. Adding more wrapper variants increases complexity for a feature with very little practical value today, so I’d prefer we keep the implementation simpler.

cipherprofessor:
Makes sense, and it's your call on scope for your own library — checked the "Go clients don't support receiving pushes either" claim too before replying: Go's stdlib `net/http` HTTP/2 implementation has no client-side push-consuming API at all (no `PushHandler` or equivalent, in current stable or on `master`) — only the server-side `Push()` method I looked at earlier. So there's genuinely no end-to-end path where this matters today, on either side.

Happy to close this PR if that's easier than leaving it open — just say the word, or close it yourself whenever's convenient. No argument from me on keeping the implementation simpler.

